PR #2, opened 2016-11-29
allow deleting of presets

jlipps:
cc @dpgraham

### DIFF
--- a/app/actions/StartServer.js
+++ b/app/actions/StartServer.js
@@ -11,6 +11,8 @@ export const SWITCH_TAB = 'SWITCH_TAB';
 export const PRESET_SAVE_REQ = 'PRESET_SAVE_REQ';
 export const PRESET_SAVE_OK = 'PRESET_SAVE_OK';
 export const GET_PRESETS = 'GET_PRESETS';
+export const PRESET_DELETE_REQ = 'PRESET_DELETE_REQ';
+export const PRESET_DELETE_OK = 'PRESET_DELETE_OK';
 
 export function startServer (evt) {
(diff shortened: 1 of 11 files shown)


app/components/StartServer/DeletePresetButton.js
```diff
@@ -0,0 +1,26 @@
...
+    const {presetDeleting, deletePreset} = this.props;
+
+    return (
+      <div>
+        <Button className={styles.startButton} type="button"
+         ptStyle={presetDeleting ? "disabled" : "warning"}
+         text={presetDeleting ? "Deleting..." : "Delete Preset"}
+         onClick={deletePreset}
+        />
+        <input type="submit" hidden={true} />
```
dpgraham: What does the hidden submit button do?

app/actions/StartServer.js
```diff
@@ -76,8 +78,20 @@ export function getPresets () {
...
     settings.get('presets').then((presets) => {
       dispatch({type: GET_PRESETS, presets});
-    }).catch((e) => {
-      console.error(e);
-    });
+    }).catch(console.error);
+  };
+}
+
+export function deletePreset (name) {
```
dpgraham: Can we use 'await' here?

app/components/StartServer/DeletePresetButton.js
jlipps: It allows this to be part of a form and receive form submit events via the enter key. But I think it probably doesn't make sense to have on this particular button since it's not part of a form. Will remove.

PR #4, opened 2016-11-30
New session

dpgraham:
* Opens up a new window and navigates it to /session
* That window has a form for filling out desired capabilities
* Submitting that form sends the caps back to the main process
* Main process starts a new session and announces back to the session window if it was successful or not

NEXT STEPS:
* Make it so sessions can be killed (either by closing window or pushing a button)
* Make it so we can open up multiple sessions


jlipps:
awesome, will do an in depth review soon

PR #5, opened 2016-12-06
Added sessions

dpgraham:
-Added button to server log page to start a new session
-Opens a dialog that has a form for adding dcaps
-Can choose server (local, remote, sauce)
-Starting a session sends message to main process at appium.js and starts an appium session using the WD client


app/components/Session.js
```diff
@@ -0,0 +1,78 @@
+import React, { Component } from 'react';
+import NewSessionForm from './Session/NewSessionForm';
+import SavedSessions from './Session/SavedSessions';
+import { Tabs, Form, Input, Button, Spin } from 'antd';
+import { ServerTypes } from '../actions/Session';
+import './Session.css';
+
+const { TabPane } = Tabs;
```
jlipps: for things other than import/exports, we don't use spaces around braces. so this should be:

```js
const {TabPane} = Tabs;
```

(we do things this way to distinguish semantically between imports/exports and simple destructured assignment)

app/components/Session.js
```diff
@@ -0,0 +1,78 @@
...
+    this.props.setSavedServerParams();
+    this.props.setLocalServerParams();
+  }
+
+  render () {
+    const { newSessionBegan, savedSessions, tabKey, switchTabs, changeServerType, serverType, setServerParam, server, 
+      requestSaveAsModal, newSession, caps, capsUUID, saveSession, isCapsDirty, sessionLoading } = this.props;
+
+    return (<Spin spinning={!!sessionLoading}>
+      <div style={{width: '100%', padding: '1em'}}>
```
jlipps: can we use css files instead of inline?

app/components/Session.js
```diff
@@ -0,0 +1,78 @@
...
+            <Form>
+              <FormItem>
+                <Input placeholder='Hostname' value={server.remote.hostname} onChange={(e) => setServerParam('hostname', e.target.value)} />
+              </FormItem>
+              <FormItem>
+                <Input placeholder='Port' value={server.remote.port} onChange={(e) => setServerParam('port', e.target.value)} />
+              </FormItem>
+            </Form>
+          </TabPane>
+          <TabPane tab='SauceLabs Server' key={ServerTypes.sauce}>
```
jlipps: let's say `Sauce Labs` instead

app/components/Session.js
```diff
@@ -0,0 +1,78 @@
...
+                <Input placeholder='Sauce Username' value={server.sauce.username} onChange={(e) => setServerParam('username', e.target.value)} />
+              </FormItem>
+              <FormItem>
+                <Input type='password' placeholder='Sauce Access Key' value={server.sauce.accessKey} onChange={(e) => setServerParam('accessKey', e.target.value)} />
+              </FormItem>
+            </Form>
+          </TabPane>
+        </Tabs>
+
+        {newSessionBegan ?
```
jlipps: with so many elements here, the conditional is a bit hard to read. what about this pattern instead?

```js
{newSessionBegan &&
  <div>...</div>
}{!newSessionBegan &&
  <div>...</div>
}
```

just trying to think of a way to separate out the two cases more clearly

app/components/Session.js
```diff
@@ -0,0 +1,78 @@
...
+        </div> : 
+        <Tabs activeKey={tabKey} onChange={(key) => switchTabs(key)}>
+          <TabPane tab='Start New Session' key='new'>
+            <NewSessionForm {...this.props} />
+          </TabPane>
+          <TabPane tab={`Saved Sessions (${savedSessions.length})`} key='saved'>
+              <SavedSessions {...this.props} />
+          </TabPane>
+        </Tabs>}
+        <div style={{float: 'right'}}>
```
jlipps: if we can let's not use inline css

app/components/Session/FormattedCaps.js
```diff
@@ -0,0 +1,20 @@
+import React, { Component } from 'react';
+import formatJSON from 'format-json';
+
+export default class NewSessionForm extends Component {
+
+  getCapsObject (caps) {
```
jlipps: this could be rewritten as i mentioned somewhere else as:

```js
let capObjs = caps.filter((cap) => !!cap.name).map((cap) => {[cap.name]: cap.value}));
return Object.assign({}, ...capObjs);
```

app/components/Session/FormattedCaps.js
```diff
@@ -0,0 +1,20 @@
...
+    caps.forEach((cap) => cap.name ? capsObject[cap.name] = cap.value : '');
+    return capsObject;
+  }
+
+  getFormattedJSON (caps) {
+    return formatJSON.plain(this.getCapsObject(caps));
+  }
+
+  render () {
+    const { caps } = this.props;
```
jlipps: no spaces around braces

app/components/Session/FormattedCaps.js
```diff
@@ -0,0 +1,20 @@
...
+    return caps && <textarea rows={15} style={{padding: '4px', width: '100%', backgroundColor: '#d3d3d3'}} onChange={() => {}} value={ this.getFormattedJSON(caps) } />
```
jlipps: inline css, also no spaces around braces

app/components/Session/Inspector.js
```diff
@@ -0,0 +1,58 @@
+/*import React, { Component } from 'react';
```
jlipps: if this is commented out, let's leave it out of this PR

app/components/Session/NewSessionForm.js
```diff
@@ -0,0 +1,90 @@
+import React, { Component } from 'react';
+import { Button, Switch, Input, Modal, Form, Icon, Row, Col } from 'antd';
+import { remote } from 'electron';
+import FormattedCaps from './FormattedCaps';
+
+const { dialog } = remote;
```
jlipps: spaces

app/components/Session/NewSessionForm.js
```diff
@@ -0,0 +1,90 @@
...
+
+  getLocalFilePath (success) {
+    dialog.showOpenDialog((filepath) => {
+      if (filepath) {
+        success(filepath);
+      }
+    });
+  }
+
+  getCapsObject () {
```
jlipps: this looks like a reimplementation of `getCapsObject` above, maybe we should reuse instead?

app/components/Session/NewSessionForm.js
```diff
@@ -0,0 +1,90 @@
...
+    const { caps } = this.props;
+    let capsObject = {};
+    caps.forEach((cap) => capsObject[cap.name] = cap.value);
+    return capsObject;
+  }
+
+  render () {
+    const { setCapabilityParam, caps, addCapability, removeCapability, saveSession, hideSaveAsModal, saveAsText, showSaveAsModal, setSaveAsText } = this.props;
```
jlipps: spaces

app/components/Session/NewSessionForm.js
```diff
@@ -0,0 +1,90 @@
...
+                  <option value='text'>text</option>
+                  <option value='boolean'>boolean</option>
+                  <option value='number'>number</option>
+                  <option value='json_object'>JSON object</option>
+                  <option value='file'>filepath</option>
+                </select>
+              </FormItem>
+              <FormItem>
+                {
+                  function () {
```
jlipps: feels like it would be cleaner to just factor this out and call it?

```js
<FormItem>
  {this.getCapControl(cap)}
</FormItem>
```

(might need to also pass in your `buttonAfter`, or build that inside this helper method instead

app/components/Session/NewSessionForm.js
```diff
@@ -0,0 +1,90 @@
...
+                      case 'text': return <Input placeholder='Value' value={cap.value} onChange={(e) => setCapabilityParam(index, 'value', e.target.value)}/>;
+                      case 'boolean': return <Switch checkedChildren={'true'} unCheckedChildren={'false'} 
+                        placeholder='Value' value={cap.value} onChange={(e) => setCapabilityParam(index, 'value', e.target.value)}/>;
+                      case 'number': return <Input placeholder='Value' value={cap.value} onChange={(e) => setCapabilityParam(index, 'value', e.target.value)}/>; 
+                      case 'json_object': return <Input type='textarea' rows={4} placeholder='Value' value={cap.value} 
+                        onChange={(e) => setCapabilityParam(index, 'value', e.target.value)}/>;
+                      case 'file': return <div>
+                        <Input placeholder='Value' value={cap.value} addonAfter={buttonAfter}/>
+                      </div>;
+                      default: break;
```
jlipps: I'd almost rather this `throw` than `break` since it's a programming error if we ever get to the default case

app/components/Session/SavedSessions.js
```diff
@@ -0,0 +1,57 @@
...
+
+
+  render () {
+    const {savedSessions, setCaps, capsUUID, switchTabs} = this.props;
+
+    return (<Row>
+      <Col span={12}>
+        <table className={SessionCSS['saved-sessions']}>
+          <tbody>
+            {savedSessions && [...savedSessions].reverse().map((session, index) => {
```
jlipps: clever way of making a copy of an array 👍

app/components/Session/SavedSessions.js
```diff
@@ -0,0 +1,57 @@
...
+              index = savedSessions.length - index - 1;
+              let anchorStyle = {display: 'block'};
+              return <tr key={index} className={session.uuid === capsUUID ? SessionCSS['active-session'] : ''} onClick={() => setCaps(session.caps, session.uuid)}> 
+                <td>
+                  <p style={anchorStyle}>
+                  {session.name || 'Untitled'}
+                  &nbsp;&nbsp;
+                  {`(${moment(session.date).format('YYYY-MM-DD')})`}
```
jlipps: what about just

```
({moment(session.date).format('YYYY-MM-DD')})
```

?

app/components/Session/SavedSessions.js
```diff
@@ -0,0 +1,57 @@
...
+                  </p>
+                </td>
+                <td>
+                  {
+                    session.uuid === capsUUID &&
+                    <div style={{float: 'right'}}>
```
jlipps: inline css in this block?

app/reducers/Session.js
```diff
@@ -0,0 +1,152 @@
...
+import { NEW_SESSION_REQUESTED, NEW_SESSION_BEGAN, NEW_SESSION_DONE,
+        SAVE_SESSION_REQUESTED, SAVE_SESSION_DONE, GET_SAVED_SESSIONS_REQUESTED, GET_SAVED_SESSIONS_DONE,
+        SESSION_LOADING, SESSION_LOADING_DONE,
+        SET_CAPABILITY_PARAM, ADD_CAPABILITY, REMOVE_CAPABILITY, SET_CAPS,
+        SWITCHED_TABS, SAVE_AS_MODAL_REQUESTED, HIDE_SAVE_AS_MODAL_REQUESTED, SET_SAVE_AS_TEXT,
+        DELETE_SAVED_SESSION_REQUESTED, DELETE_SAVED_SESSION_DONE, 
+        CHANGE_SERVER_TYPE, SET_SERVER_PARAM, SET_SERVER,
+        ServerTypes } from '../actions/Session';
+
+// Make sure there's always at least one cap
```
jlipps: maybe move this comment above the `caps` line where it is referring?

app/reducers/Session.js
```diff
@@ -0,0 +1,152 @@
...
+    remote: {},
+    sauce: {},
+  },
+  caps: [{
+    type: 'text',
+  }],
+  isCapsDirty: true,
+};
+
+export default function session (state=INITIAL_STATE, action) {
```
jlipps: i think our style is `(state = INITIAL_STATE` (spacing)

app/reducers/Session.js
```diff
@@ -0,0 +1,152 @@
...
+  switch (action.type) {
+    case NEW_SESSION_REQUESTED:
+      return Immutable.fromJS(state)
```
jlipps: can you motivate the decision to transition to immutable.js instead of sticking with the existing convention?

appium.js
```diff
@@ -35,7 +52,8 @@ function connectStartServer (win) {
 
     try {
       // set up the appium server running in this thread
-      server = await appiumServer(args, true);
+      serverArgs = args;
```
jlipps: what's the point of aliasing `serverArgs` here?

appium.js
```diff
@@ -73,12 +91,91 @@ function connectGetDefaultArgs () {
...
+    // When you close the session window, kill the associated Appium session
+    let sessionWinID = sessionWin.webContents.id;
+    sessionWin.on('closed', () => {
+      killSession(sessionWinID);
+      sessionWin = null;
+    });
+
+    // When the main window is closed, terminate the appium session and close the session window
+    win.once('closed', () => {
+      // sessionWin.close();
```
jlipps: does this not work?

appium.js
```diff
@@ -1,15 +1,32 @@
...
 var batchedLogs = [];
 
+let sessionDrivers = {};
+
+/**
+ * Kill session associated with session browser window
+ */
+function killSession (sessionWinID) {
+  if (sessionDrivers[sessionWinID]) {
+    sessionDrivers[sessionWinID].closeApp();
```
jlipps: instead of calling `closeApp` and `end`, we should be calling `quit`. and it should be `await`ed

appium.js
```diff
@@ -73,12 +91,91 @@ function connectGetDefaultArgs () {
...
+  ipcMain.on('start-session', () => {
+    let sessionWin = new BrowserWindow({width: 800, height: 600, webPreferences: {devTools: true}});
+    let sessionHTMLPath = path.resolve(__dirname, 'app', 'index.html#/session');
+    sessionWin.loadURL(`file://${sessionHTMLPath}`);
+    sessionWin.show();
+
+    // When you close the session window, kill the associated Appium session
+    let sessionWinID = sessionWin.webContents.id;
+    sessionWin.on('closed', () => {
+      killSession(sessionWinID);
```
jlipps: i imagine this should be an async function

jlipps: probably should show a spinner or something to the user while we're quitting the session

appium.js
```diff
@@ -73,12 +91,91 @@ function connectGetDefaultArgs () {
...
+
+    try {
+      let p = driver.init(desiredCapabilities);
+      event.sender.send('appium-new-session-successful');
+      await p;
+      event.sender.send('appium-new-session-ready');
+
+      // Listen for command requests
+      ipcMain.on('appium-client-command-request', async () => {
+        let type = 'source';
```
jlipps: presumably you're in the process of building this out to accommodate other request types?

appium.js
```diff
@@ -73,12 +91,91 @@ function connectGetDefaultArgs () {
...
+    });
+
+    try {
+      let p = driver.init(desiredCapabilities);
+      event.sender.send('appium-new-session-successful');
+      await p;
+      event.sender.send('appium-new-session-ready');
+
+      // Listen for command requests
+      ipcMain.on('appium-client-command-request', async () => {
```
jlipps: this `.on` section makes the `try` block longer than i'd like. can we add a `return` to the catch block then attach this handler after the `try/catch`?

package.json
```diff
@@ -41,10 +41,13 @@
   "dependencies": {
     "ansi-to-html": "0.x",
     "appium": "1.6.0",
+    "appium-base-driver": "^2.1.0",
```
jlipps: you've added this dep but i dont see it used anywhere

app/actions/Session.js
```diff
@@ -0,0 +1,284 @@
...
+const SESSION_SERVER_PARAMS = 'SESSION_SERVER_PARAMS';
+const SESSION_SERVER_TYPE = 'SESSION_SERVER_TYPE';
+
+export const ServerTypes = {
+  local: 'local',
+  remote: 'remote',
+  sauce: 'sauce',
+};
+
+function getCapsObject (caps) {
```
dpgraham: I switched it to

```javascript
Object.assign({}, ...(caps.map((cap) => { 
    return {[cap.name]: cap.value}; 
  })));
```

It confuses the opening and clothing curly braces as being the curly braces for a function and not a returned object

jlipps: ah yeah, makes sense!

dpgraham: #527aecf041cd8a53d2e332388519ffd6741db889

app/actions/Session.js
```diff
@@ -0,0 +1,284 @@
...
+  let capsObject = {};
+  caps.forEach((cap) => capsObject[cap.name] = cap.value);
+  return capsObject;
+}
+
+/**
+ * Change the caps object and then go back to the new session tab
+ */
+export function setCaps (caps, uuid) {
+  return async (dispatch) => {
```
dpgraham: 527aecf041cd8a53d2e332388519ffd6741db889

app/actions/Session.js
```diff
@@ -0,0 +1,284 @@
...
+        break;
+    }
+
+    // Start the session
+    ipcRenderer.send('appium-create-new-session', {desiredCapabilities, host, port, username, accessKey, https});
+
+    dispatch({type: SESSION_LOADING});
+
+    // If it failed, show an alert saying it failed
+    ipcRenderer.once('appium-new-session-failed', () => {
```
dpgraham: 527aecf041cd8a53d2e332388519ffd6741db889

app/actions/Session.js
```diff
@@ -0,0 +1,284 @@
...
+  };
+}
+
+
+/**
+ * Saves the caps
+ */
+export function saveSession (caps, params) {
+  return async (dispatch) => {
+    let { name, uuid } = params;
```
dpgraham: 527aecf041cd8a53d2e332388519ffd6741db889

app/actions/Session.js
```diff
@@ -0,0 +1,284 @@
...
+        date: +(new Date()),
+        name,
+        uuid,
+        caps,
+      };
+      savedSessions.push(newSavedSession);
+    } else {
+
+      // If it's an existing session, overwrite it
+      savedSessions.forEach((session, index) => {
```
dpgraham: 527aecf041cd8a53d2e332388519ffd6741db889

app/components/Session.js
dpgraham: cadddd72b0ec8ee58bc404f8add01f8e242d0858

app/components/Session.js
dpgraham: ae4dcbf761a5de492a886b2f45c91479bdb0f3e9

app/components/Session.js
dpgraham: ae4dcbf761a5de492a886b2f45c91479bdb0f3e9

app/components/Inspector.js
```diff
@@ -0,0 +1,31 @@
+import React, { Component, PropTypes } from 'react';
+import styles from './Session.css';
+
+export default class Inspector extends Component {
+  constructor (props) {
+    super(props);
+    this.state = {
```
dpgraham: 9680dbd47b20c9c9011720f01921d65125ac9d75

app/components/Inspector.js
```diff
@@ -0,0 +1,31 @@
...
+      
+    };
+  }
+
+  static propTypes = {
+    newSession: PropTypes.func,
+  }
+
+  componentWillUpdate () {
```
dpgraham: 9680dbd47b20c9c9011720f01921d65125ac9d75

app/actions/Session.js
```diff
@@ -0,0 +1,284 @@
...
+        if (session.uuid === uuid) {
+          savedSessions[index].caps = caps;
+        }
+      });
+    }
+    await settings.set(SAVED_SESSIONS, savedSessions);
+    dispatch({type: SET_CAPS, caps, uuid});
+    dispatch({type: SAVE_SESSION_DONE});
+    getSavedSessions()(dispatch);
```
dpgraham: 2bdeb214ec4c8704f6416c4d5ce673b19ee7ae5e

app/components/Session/FormattedCaps.js
dpgraham: d379ecfc254bc2a5baa11b47214e3bd5b28ea396

app/components/Session/FormattedCaps.js
dpgraham: d379ecfc254bc2a5baa11b47214e3bd5b28ea396

app/components/Session/Inspector.js
dpgraham: 3dfa5cac0b71030e986b401c8c5d94581895fab3

app/components/Session/NewSessionForm.js
dpgraham: 3dfa5cac0b71030e986b401c8c5d94581895fab3

app/components/Session/NewSessionForm.js
dpgraham: 3dfa5cac0b71030e986b401c8c5d94581895fab3

app/components/Session/SavedSessions.js
dpgraham: 5d74975f89e278d32acdea492b5f2ab52df3d494

app/components/Session/SavedSessions.js
dpgraham: 5d74975f89e278d32acdea492b5f2ab52df3d494

app/reducers/Session.js
dpgraham: My original motivations

1) **Cleaner Code** The ES6 conventions worked great for shallow copying but when deep copying was required the code became messy. I also like how Immutable operations sententially describe how the state is being changed.

e.g.)

```javascript
// With immutable.js
case SET_CAPABILITY_PARAM:
  return Immutable.fromJS(state)
    .setIn(['caps', action.index, action.name], action.value)
    .set('isCapsDirty', true)
    .toJS();
```

```javascript
// Without immutable
case SET_CAPABILITY_PARAM:
  let caps = {...nextState.caps};
  caps[action.index][action.name] = action.value;
  return {
    ...state,
    ...caps,
    isCapsDirty: true,
  }
```

2) **Enforced immutability** Using the standard ES6 convention I had mistakenly forgotten to deep copy elements which broke the React components (React expects a completely new state). I was able to spot them and fix them but it took some time and there's a risk that myself or another dev may make this mistake in the future without realizing it. ImmutableJS eliminates this class of risks altogether.

On second thought though...

I think it may be better to make use of 'combineReducers' (http://redux.js.org/docs/api/combineReducers.html) so that all data is shallow

app/reducers/Session.js
dpgraham: abd74413b623cd5f049703aab6c8b26b224d7c43

app/reducers/Session.js
dpgraham: abd74413b623cd5f049703aab6c8b26b224d7c43

appium.js
dpgraham: The way that a session is killed is by closing the window. 'appium.js' subscribes to the window closing event and then calls 'killSession' so the window will be gone before they can see a loading indicator.

Also, you can't use UI indicators in 'appium.js' because it runs on the main process.

appium.js
dpgraham: I have another branch that I'm working on that fixes this problem. I'll just remove it from this PR for now and it'll be in a future PR.

appium.js
dpgraham: This code is actually broken. It starts a new listener every time a new session is created and then when 'appium-client-command-request' is triggered by just one session window, it sends the response back to all session windows.

I'll take this out of the current PR, because it's not even being used currently. The next PR will have this all fixed.

appium.js
dpgraham: 87df2ad54e00f24585ea407aa437a461605e99d8

app/components/Session.js
dpgraham: cadddd72b0ec8ee58bc404f8add01f8e242d0858

app/components/Session/FormattedCaps.js
dpgraham: cadddd72b0ec8ee58bc404f8add01f8e242d0858

app/components/Session/NewSessionForm.js
dpgraham: cadddd72b0ec8ee58bc404f8add01f8e242d0858

dpgraham:
Changes are complete

package.json
dpgraham: 89a8054e74b34d432dd1543cad55bf091956d8f1

app/actions/Session.js
```diff
@@ -0,0 +1,285 @@
...
+  };
+}
+
+/**
+ * Set the local server hostname and port to whatever was saved in 'actions/StartServer.js' so that it
+ * defaults to what the currently running appium server is
+ */
+export function setLocalServerParams () {
+  return async (dispatch) => {
+    let port = await settings.get('SERVER_PORT');
```
jlipps: pretty sure that this will result in an unhandled rejection if anything inside this `async` function throws. not sure if we care.

app/components/Session.js
```diff
@@ -0,0 +1,79 @@
...
+              </FormItem>
+            </Form>
+          </TabPane>
+        </Tabs>
+
+        {newSessionBegan && <div key={2}>
+          <p>Session In Progress</p>
+        </div>}
+        
+        {!newSessionBegan && <Tabs activeKey={tabKey} onChange={(key) => switchTabs(key)}>
```
jlipps: is `(key) => switchTabs(key)` any different than `switchTabs`?

app/components/Session.js
```diff
@@ -0,0 +1,79 @@
...
+          <TabPane tab='Start New Session' key='new'>
+            <NewSessionForm {...this.props} />
+          </TabPane>
+          <TabPane tab={`Saved Sessions (${savedSessions.length})`} key='saved'>
+              <SavedSessions {...this.props} />
+          </TabPane>
+        </Tabs>}
+        <div className={SessionStyles['pull-right']}>
+          { capsUUID && <Button type="primary" onClick={() => saveSession(caps, {uuid: capsUUID})} disabled={!isCapsDirty}>Save</Button> }
+          <Button type="button" onClick={requestSaveAsModal} style={{marginLeft: '1em'}}>Save As</Button>
```
jlipps: still some inline css here

app/components/Session/NewSessionForm.js
```diff
@@ -0,0 +1,89 @@
...
+  render () {
+    const {setCapabilityParam, caps, addCapability, removeCapability, saveSession, hideSaveAsModal, saveAsText, showSaveAsModal, setSaveAsText} = this.props;
+
+    return <div>
+      <h4>Desired Capabilities</h4> 
+      <Row> 
+      <Col span={12}>
+        <Form inline>
+          {caps.map((cap, index) => {
+            return <div key={index} style={{marginBottom: '1em'}}>
```
jlipps: inline css

app/components/Session/NewSessionForm.js
```diff
@@ -0,0 +1,89 @@
...
+                  <option value='json_object'>JSON object</option>
+                  <option value='file'>filepath</option>
+                </select>
+              </FormItem>
+              <FormItem>
+                {this.getCapsControl(cap, index)}
+              </FormItem>
+              <FormItem>
+                { (caps.length > 1) && <Button icon='delete' onClick={() => removeCapability(index)}/> }
+                { (index === caps.length - 1) && <Button icon='plus' onClick={addCapability} style={{marginLeft: '4px'}}/> }
```
jlipps: can probably handle this in css using `::last-child` or something

appium.js
```diff
@@ -73,12 +88,74 @@ function connectGetDefaultArgs () {
...
+  ipcMain.on('start-session', () => {
+    let sessionWin = new BrowserWindow({width: 800, height: 600, webPreferences: {devTools: true}});
+    let sessionHTMLPath = path.resolve(__dirname, 'app', 'index.html#/session');
+    sessionWin.loadURL(`file://${sessionHTMLPath}`);
+    sessionWin.show();
+
+    // When you close the session window, kill the associated Appium session
+    let sessionWinID = sessionWin.webContents.id;
+    sessionWin.on('closed', async () => {
+      await killSession(sessionWinID);
```
jlipps: should probably at least log an error here if one happens during session kill

package.json
```diff
@@ -45,6 +45,7 @@
     "electron-debug": "1.x",
     "electron-settings": "2.x",
     "font-awesome": "4.x",
+    "format-json": "^1.0.3",
```
jlipps: let's use `"1.x"`

package.json
```diff
@@ -59,9 +60,11 @@
     "redux-thunk": "2.x",
     "source-map-support": "0.x",
     "teen_process": "1.x",
-    "temp": "0.x"
+    "temp": "0.x",
+    "wd": "^1.0.0"
```
jlipps: same here

package.json
```diff
@@ -59,9 +60,11 @@
...
   },
   "devDependencies": {
+    "antd": "^2.5.0",
```
jlipps: curious why `antd` is only a devDependency?

also, "2.x"

app/actions/Session.js
dpgraham: If the exception isn't caught, it just won't set the local server host and port which I think i is acceptable.

jlipps: If the exception isn't caught, it will generate an unhandled rejection warning, which in later versions of Node 7 will kill the Node process. That would be a bad user experience :-)

package.json
```diff
@@ -61,10 +61,10 @@
     "source-map-support": "0.x",
     "teen_process": "1.x",
     "temp": "0.x",
-    "wd": "^1.0.0"
+    "wd": "1.x",
+    "antd": "^2.5.0"
```
jlipps: "2.x"

app/actions/Session.js
dpgraham: This is in the web process.

jlipps: hmm hmm you're right. In that case it shouldn't be a big deal.

dpgraham:
Changes p ushed

PR #7, opened 2016-12-06
display timestamps if logTimestamps is set

jlipps:
since we get logs from Appium before it has a chance to

review please @dpgraham


dpgraham:
LGTM

PR #10, opened 2016-12-16
Added the inspector component

dpgraham:
* Fetches the source and screenshot from the running session
* Renders screenshot as a data-uri img tag
* Renders invisible divs overtop of all the images that are made visible when they're hovered over or selected
* Shows the source as a tree on the right
* Added functionality to request client methods on main process and main process sends back result
* Client calls .source, .takeScreenshot, .back, .tap, .sendKeys
* Shows a pane that has detailed information about the selected element


slipy12:
`npm run dev` returned me that
https://gist.github.com/slipy12/13d649648c01e68508512b13c3066065

Am I missing something? (it's second time i got those errors, also posted as issue with master)

dpgraham:
@slipy12 Did you call 'npm install'? If yes, then try doing 'rm -rf node_modules/ && npm install', it looks like you're missing the newest dependencies.

app/components/Inspector.js
```diff
@@ -1,13 +1,32 @@
...
+import Screenshot from './Inspector/Screenshot';
+import SelectedElement from './Inspector/SelectedElement';
+import Source from './Inspector/Source';
+import InspectorStyles from './Inspector.css';
 
 export default class Inspector extends Component {
 
+  componentWillMount () {
+    this.props.applyClientMethod({methodName: 'source'});
+    this.props.bindAppium();
```
jlipps: could there ever be a race condition in which we haven't bound the listener even though the source command might have tried to trigger the event?

app/components/Inspector/HighlighterRect.js
```diff
@@ -0,0 +1,46 @@
...
+
+      // Calculate left, top, width and height coordinates
+      let left = x1 / scaleRatio;
+      let top = y1 / scaleRatio;
+      let width = (x2 - x1) / scaleRatio;
+      let height = (y2 - y1) / scaleRatio;
+
+      // Add class + special classes to hovered and selected elements
+      let highlighterClasses = [InspectorCSS['highlighter-box']];
+      hoveredPath === node.path && highlighterClasses.push(InspectorCSS['hovered-element-box']);
```
jlipps: I'm unclear why you're reassigning `hoveredPath` and `selectedPath` here.

app/components/Inspector/Screenshot.js
```diff
@@ -0,0 +1,69 @@
...
+ * Shows screenshot of running application and divs that highlight the elements' bounding boxes
+ */
+export default class Screenshot extends Component {
+
+  constructor (props) {
+    super(props);
+    this.state = {
+      scaleRatio: 1
+    };
+    this.updateScaleRatio = debounce(this.updateScaleRatio.bind(this), 1000);
```
jlipps: 👍  on `debounce`

app/components/Inspector/Screenshot.js
```diff
@@ -0,0 +1,69 @@
...
+
+    let recursive = (node, zIndex = 0) => {
+      if (!node) return;
+      highlighterRects.push(<HighlighterRect {...this.props} 
+        node={node} 
+        zIndex={zIndex} 
+        scaleRatio={scaleRatio} 
+        key={node.path} 
+      />);
+      node.children.forEach((childNode) => recursive(childNode, zIndex + 1));
```
jlipps: mind if we just replace with

```js
for (childNode of node.children) {
  recursive(childNode, zIndex + 1);
}
```

appium.js
```diff
@@ -95,41 +96,60 @@ function connectGetDefaultArgs () {
...
-  ipcMain.on('start-session', () => {
-    let sessionWin = new BrowserWindow({width: 800, height: 600, webPreferences: {devTools: true}});
+/**
+ * Opens a new window for creating new sessions
+ */
+function connectCreateNewSessionWindow (win) {
+  ipcMain.on('create-new-session-window', () => {
+
+    // Create and open the Browser Window
+    let sessionWin = new BrowserWindow({width: 1500, height: 1000, webPreferences: {devTools: true}});
```
jlipps: NIT: not sure we want to default to something so big

appium.js
```diff
@@ -95,41 +96,60 @@ function connectGetDefaultArgs () {
...
     let sessionHTMLPath = path.resolve(__dirname, 'app', 'index.html#/session');
     sessionWin.loadURL(`file://${sessionHTMLPath}`);
     sessionWin.show();
 
-    // When you close the session window, kill the associated Appium session
-    let sessionWinID = sessionWin.webContents.id;
-    sessionWin.on('closed', async () => {
-      await killSession(sessionWinID);
+    // When you close the session window, kill it's associated Appium session (if there is one)
```
jlipps: NIT: `its`

appium.js
```diff
@@ -95,41 +96,60 @@ function connectGetDefaultArgs () {
...
+    let sessionID = sessionWin.webContents.id;
+    sessionWin.on('closed', () => {
+      if (sessionDrivers[sessionID]) {
+        sessionDrivers[sessionID].quit();
```
jlipps: shouldn't `driver.quit` be awaited?

appium.js
```diff
@@ -138,31 +158,76 @@ function connectCreateNewSession () {
...
-      event.sender.send('appium-new-session-failed', JSON.parse(e.data).value.message);
+      await killSession(event.sender);
+      event.sender.send('appium-new-session-failed');
     }
 
   });
 }
 
+/**
+ * When a Session Window makes method request, find it's corresponding driver, execute requested method
```
jlipps: NIT: `its`

appium.js
```diff
@@ -138,31 +158,76 @@ function connectCreateNewSession () {
...
+ */
+function connectClientMethodListener () {
+  ipcMain.on('appium-client-command-request', async (evt, data) => {
+    const {methodName, args = [], xpath} = data;
+    let driver = sessionDrivers[evt.sender.id];
+
+    let source, screenshot;
+    try {
+      if (methodName === 'quit') {
+        evt.sender.send('appium-session-done');
```
jlipps: NIT: might be clearer if we did `let renderer = evt.sender` up top? And then `renderer.send()` here? Just a thought

appium.js
```diff
@@ -138,31 +158,76 @@ function connectCreateNewSession () {
...
+
+        // Execute the requested method
+        if (methodName !== 'source') {
+          if (xpath) {
+            await driver.elementByXPath(xpath)[methodName](...args);
+          } else {
+            await driver[methodName](...args);
+          }
+        }
+        await Bluebird.delay(500);
```
jlipps: should probably add a comment as to why we're doing this (to allow any interaction to happen before we get the new source)

package.json
```diff
@@ -40,7 +40,9 @@
   },
   "dependencies": {
     "ansi-to-html": "0.x",
+    "antd": "2.x",
```
jlipps: I'm confused, wasn't this already a dependency?

app/components/Inspector.js
dpgraham: Good catch. I'll switch those two around.

app/components/Inspector/HighlighterRect.js
dpgraham: It's actually doing a conditional to check if it should add a class. I'm going to change it to the old fashioned 'if statement' so it's more obvious what the code's doing.

package.json
dpgraham: The previous PR still hardcoded the version. I must have forgotten to include it in the last commit.

dpgraham:
Pushed fixes

app/components/Inspector/HighlighterRect.js
jlipps: I see. Yeah it's clever, but if the only point for the assignment is to get around syntax parsing rules to make this one-line if statement work, I'd rather remove it as it's confusing.

app/actions/Inspector.js
```diff
@@ -32,11 +32,13 @@ export function bindAppium () {
         let path = (index !== undefined) && `${!parentPath ? '' : parentPath + '.'}${index}`;
 
         // Get an xpath for this element as well to use for Appium calls
-        let xpath = path && path.split('.').map((index) => `//*[${parseInt(index, 10) + 1}]`).join('');
+        let xpath = path && path.split('.').map((index) => `/*[${parseInt(index, 10) + 1}]`).join('');
```
jlipps: I'm not sure, but we may still want for the first selector to have `//` in front of it?

app/components/Inspector/HighlighterRect.js
```diff
@@ -27,8 +27,10 @@ export default class HighlighterRect extends Component {
 
       // Add class + special classes to hovered and selected elements
       let highlighterClasses = [InspectorCSS['highlighter-box']];
-      hoveredPath === node.path && highlighterClasses.push(InspectorCSS['hovered-element-box']);
-      selectedPath === node.path && highlighterClasses.push(InspectorCSS['inspected-element-box']);
+      if (hoveredPath === node.path)
```
jlipps: It's just an appium style thing but we've opted to never do brace-less ifs, with one exception:

```js
if (err) return cb(err);
```

is ok. I know this is safe but to keep with consistency, mind if we add braces?

appium.js
```diff
@@ -123,7 +123,7 @@ function connectCreateNewSessionWindow (win) {
       sessionWin = null;
     });
 
-    // If it's dev, include devTools and 'inspect element' context menu option
+    // If its dev, include devTools and 'inspect element' context menu option
```
jlipps: NIT: LOL this one needs to go `its` => `it's`. It was right before :-)

dpgraham:
Added fixes again.

jlipps:
awesome, well done!

PR #11, opened 2016-12-21
Further revisions to inspector

dpgraham:
-Encapsulated method calls on the main process in the form of a promise (the method call sends a UUID to the main process along with the methodName and args; a promise is stored with the UUID as it's key; when the method call is complete the main process reports the result back along withe the UUID; the promise is then resolved)
-Added a coordinate parser which handles determining the bounds of an iOS or Android element
-Calculates the image scaling ratio based on the width of the top element vs the calculated with of the image

### DIFF
--- a/app/actions/Inspector.js
+++ b/app/actions/Inspector.js
@@ -61,7 +61,7 @@ function xmlToJSON (source) {
     let path = (index !== undefined) && `${!parentPath ? '' : parentPath + '.'}${index}`;
 
     // Get an xpath for this element as well to use for Appium calls
-    let xpath = path && path.split('.').map((index) => `//*[${parseInt(index, 10) + 1}]`).join('');
+    let xpath = path && path.split('.').map((index) => `${path.length === 0 ? '/' : '//'}*[${parseInt(index, 10) + 1}]`).join('');
 
     // Translate attributes array to an object
     let attrObject = {};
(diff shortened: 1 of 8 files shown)


app/actions/Inspector.js
```diff
@@ -13,52 +14,79 @@ export const SET_EXPANDED_PATHS = 'SET_EXPANDED_PATHS';
...
+
+    // Get a dot separated path (root doesn't have a path)
+    let path = (index !== undefined) && `${!parentPath ? '' : parentPath + '.'}${index}`;
+
+    // Get an xpath for this element as well to use for Appium calls
+    let xpath = path && path.split('.').map((index) => `${path.length === 0 ? '/' : '//'}*[${parseInt(index, 10) + 1}]`).join('');
+
+    // Translate attributes array to an object
+    let attrObject = {};
+    [...(xmlNode.attributes || [])].forEach((attribute) => attrObject[attribute.name] = attribute.value);
```
jlipps: trying to sneak that `forEach` back in there eh? :-)

app/actions/Inspector.js
```diff
@@ -104,9 +132,17 @@ export function unselectHoveredElement (path) {
  * Requests a method call on appium 
  */
 export function applyClientMethod (params) {
-  return (dispatch) => {
-    ipcRenderer.send('appium-client-command-request', params);
-    dispatch({type: METHOD_CALL_REQUESTED});
+  return async (dispatch) => {
+    try {
+      dispatch({type: METHOD_CALL_REQUESTED});
+      let {source, screenshot, result} = await callClientMethod(params.methodName, params.args, params.xpath);
```
jlipps: yeah, this is nice 👍

PR #27, opened 2017-02-02
revamp UI of new session creation window

jlipps:
@dpgraham can you go ahead and take a look? this is all for the new session creation window. i'll be working on the inspector window and pushing it in a follow-up commit to this PR. lot of crazy CSS stuff in here that's probably easiest to evaluate by checking out the branch and building.

### DIFF
--- a/app/components/Inspector.js
+++ b/app/components/Inspector.js
@@ -1,6 +1,6 @@
 import React, { Component } from 'react';
 import ReactDOM from 'react-dom';
-import { Card } from 'antd';
+import { Card, Icon, Button, Spin } from 'antd';
 import Screenshot from './Inspector/Screenshot';
 import SelectedElement from './Inspector/SelectedElement';
 import Source from './Inspector/Source';
@@ -8,27 +8,58 @@ import InspectorStyles from './Inspector.css';
 
 export default class Inspector extends Component {
 
+  constructor () {
+    super();
+    this.didInitialResize = false;
+  }
+
   componentWillMount () {
+    const chromeHeight = 48;
+    if (!this.didInitialResize) {
+      // resize width to something sensible for using the inspector on first run
+      window.resizeTo(1150, window.clientHeight - chromeHeight);
+      this.didInitialResize = true;
+    }
     this.props.bindAppium();
     this.props.applyClientMethod({methodName: 'source'});
   }
 
   render () {
-    const {screenshot, selectedElement = {}} = this.props;
+    const {screenshot, selectedElement = {}, applyClientMethod} = this.props;
     const {path} = selectedElement;
+
+    let actions = <span>
+      <Button icon='arrow-left' onClick={() => applyClientMethod({methodName: 'back'})}>
+        Back
+      </Button>
+      <Button icon='reload' onClick={() => applyClientMethod({methodName: 'source'})}>
+        Refresh
+      </Button>
+    </span>;
 
     return <div className={InspectorStyles['inspector-container']}>
       <div className={InspectorStyles['screenshot-container']}>
-          {screenshot && <Screenshot {...this.props} />}
+        {screenshot && <Screenshot {...this.props} />}
+        {!screenshot &&
+          <Spin size="large" spinning={true}>
+            <div className={InspectorStyles.screenshotBox} />
+          </Spin>
+        }
       </div>
       <div className={InspectorStyles['source-tree-container']}>
-        <Card title='Source' className={InspectorStyles['source-tree-card']}>
+        <Card
+         title={<span><Icon type="file-text" /> App Source</span>}
+         extra={actions}
+         className={InspectorStyles['source-tree-card']}>
           <Source {...this.props} />
         </Card>
       </div>
-      <div className={InspectorStyles['source-tree-container']}>
-        <Card title='Inspector' className={InspectorStyles['source-tree-card']}>
-          {path && <SelectedElement {...this.props}/>}
+      <div className={`${InspectorStyles['source-tree-container']} ${InspectorStyles['element-detail-container']}`}>
+        <Card
+         title={<span><Icon type="tag-o" /> Selected Element</span>}
+         className={InspectorStyles['source-tree-card']}>
+         {path && <SelectedElement {...this.props}/>}
+         {!path && <i>Select an element in the source to begin.</i>}
         </Card>
       </div>
     </div>;
--- a/app/components/Inspector.js
+++ b/app/components/Inspector.js
@@ -17,7 +17,7 @@ export default class Inspector extends Component {
     const chromeHeight = 48;
     if (!this.didInitialResize) {
       // resize width to something sensible for using the inspector on first run
-      window.resizeTo(1150, window.clientHeight - chromeHeight);
+      window.resizeTo(1080, window.clientHeight - chromeHeight);
       this.didInitialResize = true;
     }
     this.props.bindAppium();
@@ -29,10 +29,10 @@ export default class Inspector extends Component {
     const {path} = selectedElement;
 
     let actions = <span>
-      <Button icon='arrow-left' onClick={() => applyClientMethod({methodName: 'back'})}>
+      <Button icon='arrow-left' onClick={() => applyClientMethod({methodName: 'back'})} size="small" >
         Back
       </Button>
-      <Button icon='reload' onClick={() => applyClientMethod({methodName: 'source'})}>
+      <Button icon='reload' onClick={() => applyClientMethod({methodName: 'source'})} size="small">
         Refresh
       </Button>
       <Button icon='close' onClick={() => quitSession()} size="small">
--- a/app/components/Inspector.js
+++ b/app/components/Inspector.js
@@ -25,7 +25,7 @@ export default class Inspector extends Component {
   }
 
   render () {
-    const {screenshot, selectedElement = {}, applyClientMethod} = this.props;
+    const {screenshot, selectedElement = {}, applyClientMethod, quitSession} = this.props;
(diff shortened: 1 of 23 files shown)


app/reducers/Session.js
```diff
@@ -112,15 +112,19 @@ export default function session (state = INITIAL_STATE, action) {
       };
 
     case DELETE_SAVED_SESSION_DONE:
-      return omit(state, 'deletingSession');
+      return {
+        ...state,
+        deletingSession: false,
+        capsUUID: null
```
dpgraham: Why is capsUUID deleted but caps isn't?

package.json
```diff
@@ -42,7 +42,7 @@
   "dependencies": {
     "ansi-to-html": "0.x",
     "antd": "2.x",
-    "appium": "1.6.x",
+    "appium": "1.6.4-beta",
```
dpgraham: We should look at adding a way for the user to choose the version of appium

app/components/Session.js
```diff
@@ -17,36 +17,38 @@ export default class Session extends Component {
...
               </FormItem>
             </Form>
           </TabPane>
-          <TabPane tab='Remote Server' key={ServerTypes.remote}>
+          <TabPane tab='Custom Server' key={ServerTypes.remote}>
             <Form>
               <FormItem>
-                <Input placeholder='Hostname' value={server.remote.hostname} onChange={(e) => setServerParam('hostname', e.target.value)} />
+                <Input placeholder='127.0.0.1' addonBefore="Remote Host" value={server.remote.hostname} onChange={(e) => setServerParam('hostname', e.target.value)} size="large" />
               </FormItem>
```
dpgraham: I like the addons, looks sharp

app/components/Session.js
```diff
@@ -17,36 +17,38 @@ export default class Session extends Component {
...
+                </Card>
               </FormItem>
             </Form>
           </TabPane>
-          <TabPane tab='Remote Server' key={ServerTypes.remote}>
+          <TabPane tab='Custom Server' key={ServerTypes.remote}>
             <Form>
               <FormItem>
-                <Input placeholder='Hostname' value={server.remote.hostname} onChange={(e) => setServerParam('hostname', e.target.value)} />
+                <Input placeholder='127.0.0.1' addonBefore="Remote Host" value={server.remote.hostname} onChange={(e) => setServerParam('hostname', e.target.value)} size="large" />
```
dpgraham: I like the addons, looks sharp

app/components/Session/NewSessionForm.js
```diff
@@ -1,10 +1,11 @@
 import React, { Component } from 'react';
-import { Button, Switch, Input, Modal, Form, Icon, Row, Col } from 'antd';
+import { Button, Switch, Input, Modal, Form, Icon, Row, Col, Select } from 'antd';
 import { remote } from 'electron';
 import FormattedCaps from './FormattedCaps';
 import SessionStyles from '../Session.css';
 const {dialog} = remote;
 const FormItem = Form.Item;
+const Option = Select.Option;
```
dpgraham: I think we should be consistent with using destructuring

```javascript
const {Item:FormItem} = Form;
const {Option} = Select;
```

app/components/Session/NewSessionForm.js
```diff
@@ -81,48 +82,51 @@ export default class NewSessionForm extends Component {
...
               </FormItem>
               <FormItem>
-                <select onChange={(e) => this.handleSetType(index, e.target.value)} value={cap.type}>
-                  <option value='text'>text</option>
-                  <option value='boolean'>boolean</option>
-                  <option value='number'>number</option>
-                  <option value='json_object'>JSON object</option>
-                  <option value='file'>filepath</option>
-                </select>
+                <Select onChange={(val) => this.handleSetType(index, val)} defaultValue={cap.type} size="large" style={{width: 120}}>
```
dpgraham: Why is this 'defaultValue' instead of 'value'?

Strangely, I tested both attributes and they both work.

app/components/Session/SavedSessions.js
```diff
@@ -1,56 +1,88 @@
...
-                </td>
-              </tr>;
-            })}
-          </tbody>
-        </table>
+        <Table className={SessionCSS['saved-sessions']}
+         pagination={false}
+         dataSource={dataSource}
+         columns={columns}
+         onRowClick={onRowClick}
```
dpgraham: I think we should also make use of the 'rowClassName' attribute. The rowClassName attribute returns the classname that the row should have,

```javascript
getRowClassName (record) {
    const {capsUUID} = this.props;
    return capsUUID === record.key ? SessionCSS.selected: '';
  }
```
rowClassName={this.getRowClassName}
```

```

app/components/StartServer/StartServer.css
```diff
@@ -5,7 +5,7 @@
     display: flex;
     align-items: center;
     justify-content: center;
-    background: linear-gradient(180deg, #f0f0f0, #e0e0e0);
+    background: linear-gradient(180deg, #f9f9f9, #f2f2f2, 5%, #f2f2f2, 95%, #d9d9d9);
```
dpgraham: I think this can be taken out. The container is inside of the root and the root has the same linear-gradient.

dpgraham:
Code review complete. It's looking good! 

I made some comments, there were a couple minor stylistic things I found.

I tested this locally and tested all of the changes that I recommended so if you want to pull those changes into the branch, lmk and I'll commit/push them.

app/reducers/Session.js
jlipps: i was just doing this to fix an exception that was thrown on delete when capsUUID was still set; the formatted-caps section was having problems with it I think. but we should probably also set caps to null!

package.json
jlipps: yeah that's a good idea. for now we'll just cut a new appium-desktop release whenever there's a new version of appium.

app/components/Session/NewSessionForm.js
jlipps: yeah, looks good

app/components/Session/NewSessionForm.js
jlipps: when I tried `value`, I found that I couldn't actually change the value in the drop-down. *shrug*

app/components/Session/SavedSessions.js
```diff
@@ -1,56 +1,88 @@
...
+          date: moment(session.date).format('YYYY-MM-DD')
+        };
+      });
+    }
+
+    let onRowClick = (record) => {
+      let session = this.sessionFromUUID(record.key);
+      setCaps(session.caps, session.uuid);
+    };
+
```
jlipps: I'm fine moving this out into a member function, but I'm not sure what binding it in the constructor would do. Aren't member functions already bound?

app/components/Session/SavedSessions.js
jlipps: yes, this is a good idea. I overlooked this and couldn't find an easy way to do the highlighting.

app/components/StartServer/StartServer.css
jlipps: yes, I was playing around with where to put this and must have left it in a random place

jlipps:
thanks for the feedback!

app/components/Session/SavedSessions.js
dpgraham: 'this' gets bound to the evt object

jlipps: I see what you're saying. I usually bind to `this` in the calling location, but I see your point that you could bind it in the constructor instead.

app/components/Inspector.js
```diff
@@ -1,34 +1,68 @@
...
+  constructor () {
+    super();
+    this.didInitialResize = false;
+  }
+
   componentWillMount () {
+    const chromeHeight = 48;
+    if (!this.didInitialResize) {
+      // resize width to something sensible for using the inspector on first run
+      window.resizeTo(1080, window.clientHeight - chromeHeight);
```
dpgraham: What's the reason for doing this here instead of in appium.js?

app/components/Inspector/Screenshot.js
```diff
@@ -46,17 +51,25 @@ export default class Screenshot extends Component {
     // Recurse through the 'source' JSON and render a highlighter rect for each element
     const highlighterRects = [];
 
+    let highlighterXOffset = 0;
+    if (this.containerEl) {
+      const screenshotEl = this.containerEl.querySelector('img');
+      highlighterXOffset = screenshotEl.getBoundingClientRect().left -
+                           this.containerEl.getBoundingClientRect().left;
+    }
+
```
dpgraham: I assume you had to do this after you changed the CSS from 100% width to 100% height?

dpgraham:
lgtm, just added a couple comments

app/components/Inspector.js
jlipps: This isn't when the BrowserWindow is created; this is when the router navigates from `#/session` to `#/inspector`. Regardless of how big the Session window was, or whether the user made it a random size, I want the inspector to start at a reasonable width (so that the screenshot and all the columns are nicely visible)

app/components/Inspector/Screenshot.js
jlipps: No, this change is because I wanted to ensure that the image was always horizontally centered in the bounding box of the overall column. Given that, the top left of the image is no longer guaranteed to be at `0, 0` in the bounding column. This we may need to offset the X position based on where the top left of the image is in the column.

jlipps:
Thanks for the comments! Will merge and get a new one up for you right away :-)

PR #28, opened 2017-02-06
reorganize project

jlipps:
I didn't like that config files and the main process files were all mixed in. So basically put all of our app code in `app`, with two subdirs, `main` and `renderer`. This required an update of some of the config.

Also went ahead and removed sqlite3 since it's no longer needed.

I think this organization is a lot cleaner!


dpgraham:
Looks good. Good call on moving 'main' and 'renderer' into separate directories.

PR #39, opened 2017-02-23
enable launching of new session window even when server isn't running

jlipps:
review please @dpgraham

### DIFF
--- a/app/main/appium.js
+++ b/app/main/appium.js
@@ -111,58 +111,61 @@ function connectGetDefaultArgs () {
  */
 function connectCreateNewSessionWindow (win) {
   ipcMain.on('create-new-session-window', () => {
+    createNewSessionWindow(win);
+  });
+}
 
-    // Create and open the Browser Window
-    let sessionWin = new BrowserWindow({
-      width: 920, 
-      minWidth: 920, 
-      height: 570, 
-      minHeight: 570, 
-      title: "Start Session", 
-      backgroundColor: "#f2f2f2", 
-      webPreferences: {
-        devTools: true
-      }
-    });
-    // note that __dirname changes based on whether we're in dev or prod;
-    // in dev it's the actual dirname of the file, in prod it's the root
-    // of the project (where main.js is built), so switch accordingly
-    let sessionHTMLPath = path.resolve(__dirname,  isDev ? '..' : 'app', 'renderer', 'index.html');
-    // on Windows we'll get backslashes, but we don't want these for a browser URL, so replace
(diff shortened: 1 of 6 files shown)


dpgraham:
When running new session window from the menu it still says under the Automatic Server tab `Will use currently-running Appium Desktop server at http://localhost:4723`. We should hide this tab if there is no server running from within Appium Desktop.

dpgraham:
I added a commit that fixes the issue if you wanted to have a look @jlipps

app/main/appium.js
```diff
@@ -17,6 +18,9 @@ var batchedLogs = [];
 
 let sessionDrivers = {};
 
+// Delete saved server args, don't start until a server has been started
+settings.delete('SERVER_ARGS');
```
jlipps: wouldn't this need to be `await`ed to avoid any potential race conditions?

app/main/appium.js
```diff
@@ -60,6 +64,7 @@ function connectStartServer (win) {
     try {
       // set up the appium server running in this thread
       server = await appiumServer(args, true);
+      settings.set('SERVER_ARGS', args);
```
jlipps: same here and below too

app/renderer/actions/Session.js
```diff
@@ -313,11 +313,18 @@ export function setServerParam (name, value) {
...
-    let port = await settings.get('SERVER_PORT');
-    let host = await settings.get('SERVER_HOST');
-    dispatch({type: SET_SERVER_PARAM, serverType: ServerTypes.local, name: 'port', value: port});
-    dispatch({type: SET_SERVER_PARAM, serverType: ServerTypes.local, name: 'hostname', value: host});
+  return async (dispatch, getState) => {
+    let serverArgs = await settings.get('SERVER_ARGS');
+    if (serverArgs) {
+      dispatch({type: SET_SERVER_PARAM, serverType: ServerTypes.local, name: 'port', value: serverArgs.port});
+      dispatch({type: SET_SERVER_PARAM, serverType: ServerTypes.local, name: 'hostname', value: 'localhost'});
+    } else {
```
jlipps: maybe we should add a comment here explaining when we would run into this case?

app/main/appium.js
dpgraham: I'll make it synchronous

jlipps: if we're not going to `await` this we should at least add a `.catch` to it since it's a promise and we don't want any unhandled rejections.

dpgraham: I just changed it to '.deleteSync'

jlipps: works for me

PR #40, opened 2017-02-23
Added autoUpdater UI

dpgraham:
* Created a container that is opened in a separate Browser Window called update
* The 'Update' components interacts with the main window which checks for updates
* If an update is available, the window pops up and the user is guided through download process
* Added a mock updater that overrides autoUpdate (see https://github.com/electron-userland/electron-builder/blob/master/docs/Auto%20Update.md) so that we can test it in dev mode and not have to go through an actual update process in order to build and test the UI


dpgraham:
@jlipps Please review

PR #43, opened 2017-02-28
show version of appium server in startserver component

jlipps:
review please @dpgraham


dpgraham:
Couldn't you just cut out the middle man and import the package.json from inside the renderer window instead of importing it in the main window and sending it over to the renderer window?

dpgraham:
Looks good. I'll rebase it right now.

PR #44, opened 2017-02-28
CI env variables

dpgraham:
* Added the encrypted env variables used by Travis CI
* Improved the error logging when element not found, tells user to refresh page


dpgraham:
Changes made.

PR #58, opened 2017-03-03
fix issue with typing into desired caps box

jlipps:
there was some issue with settings.defaults not working the way i expected. i put it in its own file and now we're just using the same settings object across the board so it should be safer.

review please @dpgraham.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## [1.0.0-beta.2](https://github.com/appium/appium-desktop/compare/v1.0.0-beta.1...v1.0.0-beta.2) (2017-03-02)
+
+### Changes
+
+* Fix: issue where typing into a desired capability box was prevented if the app had not been used before. ([#57](https://github.com/appium/appium-desktop/issues/57))
+
+
 ## [1.0.0-beta.1](https://github.com/appium/appium-desktop/compare/v1.0.0-beta...v1.0.0-beta.1) (2017-03-02)
 
 ### Changes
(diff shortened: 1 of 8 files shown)


dpgraham:
Looks good to me. 

I think we need to add some coding standards with respect to reducers and how the state is transformed, otherwise this kind of thing could bite us again in the future. 

My thoughts: 

1. We should always extend the state and never overwrite it (we do this already

e.g.)
```javascript
var newState = {...state, <changes>};
```

2. The initial state should define every property up front (properties shouldn't be introduced to the state by a future action) and each property should always be the same type and never undefined or null (booleans should be false not undefined; strings should be blank '', arrays should be empty [], etc...).

e.g.)
```javascript
const initialState = {
  flag: false, // boolean
  str: '', // string
  arr: [],  // array
};

// Reducers
let newState = {...state, flag: action.flag || false};
let newState = {...state, str: action.str || ''};
let newState = {...state, arr: action.arr || []};
```

3. Objects should always be updated via the spread `...` notation (unless the case specifically requires the object to be overwritten). If an object is incorrectly overwritten, we risk having undefined references.

e.g.)
```javascript
const initialState = {
  server: {
    local: {},
    remote: {},
    sauce: {},
    testobject: {},
  }
};

//...
let newState = {
  ...state,
  server: {
    ...state.server,
    action.server,
  },
}
```

jlipps:
@dpgraham I like your approach to state sanity. Care to add it to a style.md or the end of CONTRIBUTING.md?

PR #107, opened 2017-03-28
Remove sudo from command. Only run ntpudate on Mac

dpgraham:


dpgraham:
This didn't work when I tried it on Travis. Closing.

PR #134, opened 2017-04-11
Do not quit attached session when closing the inspector

sa-spag:
Attached sessions get terminated when closing the inspector, which is quite annoying in my opinion. The initiator of the session should be responsible for deleting it.

### DIFF
--- a/app/main/appium.js
+++ b/app/main/appium.js
@@ -151,7 +151,9 @@ export function createNewSessionWindow (win) {
   sessionWin.on('closed', async () => {
     const driver = sessionDrivers[sessionID];
     if (driver) {
-      if (!driver._isAttachedSession) await driver.quit();
+      if (!driver._isAttachedSession) {
+        await driver.quit();
+      }
       delete sessionDrivers[sessionID];
     }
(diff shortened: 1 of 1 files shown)


app/main/appium.js
```diff
@@ -149,8 +149,9 @@ export function createNewSessionWindow (win) {
   // When you close the session window, kill its associated Appium session (if there is one)
   let sessionID = sessionWin.webContents.id;
   sessionWin.on('closed', async () => {
-    if (sessionDrivers[sessionID]) {
-      await sessionDrivers[sessionID].quit();
+    const driver = sessionDrivers[sessionID];
+    if (driver) {
+      if (!driver._isAttachedSession) await driver.quit();
```
jlipps: we try not to do one-line ifs in our style, can you include the braces?

sa-spag:
Should we add [`curly`](http://eslint.org/docs/rules/curly) rule to ESLint?

jlipps:
Looks like we have some level of it here: https://github.com/appium/eslint-config-appium/blob/master/index.js#L44

PR #172, opened 2017-05-05
use fix-path to ensure we get access to binaries on user's shell PATH…

### LINKED ISSUE #168: Cannot start a session
## Appium or Appium Desktop?

You are reporting an issue at the Appium Desktop repository. Appium Desktop is a wrapper around Appium. If you are having trouble running tests, it is much more likely that the problem you are encountering is not a problem with Appium Desktop but with Appium. For that reason we require you to have tried your testcase with Appium CLI before reporting issues. Check the checkbox below to confirm that you have proven your issue does _not_ reproduce on Appium itself:

- [x] I have verified that my issue does not occur with Appium and should be investigated as an Appium Desktop issue

## The problem

I cannot start a session with Appium Desktop.   

I installed carthage (0.22.0) in /usr/local/bin/.  However, when I try to start a session with Appium desktop, it says the carthage is not found.  

Then it failed when building the WebDriveAgent project.  However, I can run the same command in the terminal which succeeded without any problem.  

Here is the log: 
[startSessionInGUI.pdf](https://github.com/appium/appium/files/970817/startSessionInGUI.pdf)

Here is the log from running it in terminal:
[runWebDriveInTerminal.pdf]
(https://github.com/appium/appium/files/970823/runWebDriveInTerminal.pdf)

## Environment

- I am running Appium Desktop version 1.0.0.
- Xcode 8.3.2
- I am on (pick one):
    - [x ] Mac
    - [ ] Windows
    - [ ] Linux

jlipps:
… (fix #168)

### DIFF
--- a/app/main/main.development.js
+++ b/app/main/main.development.js
@@ -3,6 +3,7 @@ import { initializeIpc } from './appium';
 import menuTemplates from './menus';
 import path from 'path';
 import shellEnv from 'shell-env';
+import fixPath fro 'fix-path';
 
 let menu;
 let template;
@@ -21,6 +22,9 @@ if (!isDev) {
   // normally loaded in a shell, so work around with the shell-env module
   const decoratedEnv = shellEnv.sync();
   process.env = {...process.env, ...decoratedEnv};
+
+  // and we need to do the same thing with PATH
+  fixPath();
 }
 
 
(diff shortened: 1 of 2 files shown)


dpgraham:
lgtm

PR #177, opened 2017-05-10
Use env variables for SauceLabs and TO

dpgraham:
* If no sauce username or key provided, use what's in env variables
* If no TestObject host provided or TestObject key use what's in env variables

(reviewer: @jlipps)

### DIFF
--- a/app/renderer/actions/Session.js
+++ b/app/renderer/actions/Session.js
@@ -163,8 +163,8 @@ export function newSession (caps, attachSessId = null) {
       case ServerTypes.sauce:
         host = 'ondemand.saucelabs.com';
         port = 80;
-        username = session.server.sauce.username;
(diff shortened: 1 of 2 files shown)


dpgraham:
That's right, yes.

jlipps:
Just wondering about the UX of this. What about this thought: if we _are_ going to end up using the env vars (i.e., will not use something the user has typed in because they didn't type anything in, and we found environment variables with the appropriate names), can we adjust the placeholder text to reflect this? something like `Using data found in $SAUCE_ACCESS_KEY`?

dpgraham:
Yeah I think that's a good idea. If they have the env set show a placeholder. I'll make that change.

dpgraham:
Pushed a change

PR #179, opened 2017-05-10
Optimize xpath

dpgraham:
* Added unit tests to project
* Wrote helper method that takes a DOMNode and an array of unique attributes and returns an optimal xpath of the form `//tagname[@<id>=<id-value]/other-tag[<index>]/other-tag[<index>]`
* Changed the component so that it uses the xpath optimizer instead of just crawling through indices

(reviewers @jlipps)


dpgraham:
Changes made.

Xpath's that have non unique attributes now follow the form

`(//tag-name[@id="identified"])[1]`

PR #183, opened 2017-05-12
Add buttons that scroll source left and right

dpgraham:
* Added component that has a left and right scroll button that takes a container as an argument
* Clicking left makes the scrollLeft -= 20, click right makes the scrollLeft += 20
* Doesn't show component if clientWidth > scrollWidth

(reviewer @jlipps)

### DIFF
--- a/app/renderer/app.global.css
+++ b/app/renderer/app.global.css
@@ -8,6 +8,10 @@ body {
   background: linear-gradient(180deg, #f9f9f9, #f2f2f2, 5%, #f2f2f2, 95%, #d9d9d9);
 }
 
+#root .ant-spin-nested-loading > div > .ant-spin {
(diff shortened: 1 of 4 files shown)


jlipps:
All looks good to me, however I noticed that if I open the inspector in its default size, and then simply open up the xml document from there, the buttons don't show up. I actually have to trigger a resize event. Can we hook this up to check the state based on other actions in the tree? That way it can remember to check whether we need to show the buttons if we just click on a tree element that happens to expand out wide.

dpgraham:
I made it so that whenever the source is clicked on, it checks again to determine if we should see the scroll buttons or not.

In the case you're talking about, is it actually invisible in spite of the fact that the XML is too large?

PR #196, opened 2017-05-18
Gets list of running sessions from Appium

dpgraham:
* Instead of having to hardcode session id's, the desktop app gets the list of sessions from /wd/hub/sessions endpoint

attach session

(reviewer @jlipps)

### DIFF
--- a/app/renderer/actions/Session.js
+++ b/app/renderer/actions/Session.js
@@ -35,6 +35,9 @@ export const SERVER_ARGS = 'SERVER_ARGS';
 
 export const SET_ATTACH_SESS_ID = 'SET_ATTACH_SESS_ID';
 
+export const GET_SESSIONS_REQUESTED = 'GET_SESSIONS_REQUESTED';
+export const GET_SESSIONS_DONE = 'GET_SESSIONS_DONE';
+
 export const ServerTypes = {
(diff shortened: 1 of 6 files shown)


app/renderer/actions/Session.js
```diff
@@ -338,6 +342,7 @@ export function changeServerType (serverType) {
 export function setServerParam (name, value) {
   return (dispatch, getState) => {
     dispatch({type: SET_SERVER_PARAM, serverType: getState().session.serverType, name, value});
+    getRunningSessions()(dispatch, getState);
```
jlipps: this is cool as long as it doesn't happen on every keystroke

dpgraham: It does happen on every keystroke, I should debounce that.

And no we can't manually enter a session ID, I'll update that as well.

dpgraham: Just made the changes now

* It uses a combobox instead of select so that you can hardcode values
* Debounced it 500ms so it doesn't do the `getRunningSessions` on every keystroke

dpgraham:
Pushed the documentation change. Yeah we should document some of the new features and also make it part of the process to document any new features as part of the PR process.

PR #223, opened 2017-06-20
Catch source/screenshot exceptions and report error to UI

dpgraham:
* Added more try-catch blocks to appium.js thread so that if the source and screenshot fails it still continues but just reports the sourceError and screenshotError to the running browser window
* Added sourceError and screenshotError to the store
* If sourceError or screenshotError is present, show an error message in place of where source and screenshot go

### DIFF
--- a/app/main/appium.js
+++ b/app/main/appium.js
@@ -295,7 +295,7 @@ function connectClientMethodListener () {
           source = await driver.source();
         } catch (e) {
           if (e.status === 6) {
-            renderer.send('appium-session-done', e);
+            throw e;
           }
           sourceError = e;
         }
@@ -307,7 +307,7 @@ function connectClientMethodListener () {
           screenshot = await driver.takeScreenshot();
         } catch (e) {
           if (e.status === 6) {
-            renderer.send('appium-session-done', e);
+            throw e;
           }
           screenshotError = e;
         }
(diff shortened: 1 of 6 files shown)


app/main/appium.js
```diff
@@ -288,18 +288,37 @@ function connectClientMethodListener () {
...
-        screenshot = await driver.takeScreenshot();
-        renderer.send('appium-client-command-response', {source, screenshot, uuid, result});
+        // Try getting the source
+        let source;
+        let sourceError;
+        try {
+          source = await driver.source();
+        } catch (e) {
+          if (e.status === 6) {
+            renderer.send('appium-session-done', e);
```
jlipps: why not just rethrow if `e.status === 6`? then we avoid the triplicate

app/main/appium.js
```diff
@@ -288,18 +288,37 @@ function connectClientMethodListener () {
...
+        }
+
+        // Try getting the screenshot
+        let screenshot;
+        let screenshotError;
+        try {
+          screenshot = await driver.takeScreenshot();
+        } catch (e) {
+          if (e.status === 6) {
+            renderer.send('appium-session-done', e);
```
jlipps: same here

dpgraham:
Fix pushed.

PR #224, opened 2017-06-23
add basic recording functionality

jlipps:
Added action recorder and code generator


app/renderer/components/Inspector/Inspector.js
```diff
@@ -24,23 +29,70 @@ export default class Inspector extends Component {
...
+      {!isRecording &&
+       <Tooltip title="Start Recording">
+        <Button icon="eye-o" onClick={startRecording}/>
+       </Tooltip>
+      }
+      {isRecording &&
+       <Tooltip title="Pause Recording">
+         <Button icon="pause" onClick={pauseRecording}/>
+       </Tooltip>
+      }
```
dpgraham: Is there a way to reset the recording?

app/renderer/actions/Inspector.js
```diff
@@ -116,10 +128,27 @@ export function unselectHoveredElement (path) {
  * Requests a method call on appium
  */
 export function applyClientMethod (params) {
-  return async (dispatch) => {
+  return async (dispatch, getState) => {
+    let isRecording = params.methodName !== 'quit' &&
+                      params.methodName !== 'source' &&
```
dpgraham: The methodName is never set to `source`. The source method is called every time.

If we were to add `source` as a client method call, though, I think we'd want to record it.

app/renderer/client-frameworks/framework.js
```diff
@@ -0,0 +1,93 @@
+export default class Framework {
+
```
dpgraham: Could we put client-frameworks and highlighting directories into a utils or helpers directory?

app/renderer/components/Inspector/Inspector.js
```diff
@@ -24,23 +29,70 @@ export default class Inspector extends Component {
...
+           size="small" type={boilerplateType}
+          />
+        </Tooltip>
+      }
+      {!!recordedActions.length &&
+        <Tooltip title="Clear Actions">
+          <Button icon="delete" onClick={clearRecording} size="small" />
+        </Tooltip>
+      }
+      {!isRecording &&
```
dpgraham: These three instances of `!!recorderActions.length &&` can be replaced with one and you can either wrap the three components in a div or make them an array. Not a big deal if left as is though.

jlipps:
ok @dpgraham all comments addressed and ruby/python frameworks added

PR #239, opened 2017-07-04
Added locator tester modal

dpgraham:
* Added icon to the top bar that opens the modal
* Can select a locator strategy from a hard-coded select dropdown
* Can enter a locator value in an input text
* Shows the results in the same modal, inside a multi-select (only shows the ID's)
* Added a button that clicks on the selected element

screen shot 2017-07-04 at 4 16 11 pm

screen shot 2017-07-04 at 4 11 48 pm

screen shot 2017-07-04 at 4 12 36 pm

screen shot 2017-07-04 at 4 15 40 pm

### DIFF
--- a/app/renderer/actions/Inspector.js
+++ b/app/renderer/actions/Inspector.js
@@ -32,6 +32,12 @@ export const SAVED_FRAMEWORK = 'SAVED_FRAMEWORK';
 export const RECORD_ACTION = 'RECORD_ACTION';
 export const SET_SHOW_BOILERPLATE = 'SET_SHOW_BOILERPLATE';
 
+export const SHOW_LOCATOR_TEST_MODAL = 'SHOW_LOCATOR_TEST_MODAL';
+export const HIDE_LOCATOR_TEST_MODAL = 'HIDE_LOCATOR_TEST_MODAL';
+export const SET_LOCATOR_TEST_STRATEGY = 'SET_LOCATOR_TEST_STRATEGY';
+export const SET_LOCATOR_TEST_VALUE = 'SET_LOCATOR_TEST_VALUE';
+export const SEARCHING_FOR_ELEMENTS = 'SEARCHING_FOR_ELEMENTS';
+export const SEARCHING_FOR_ELEMENTS_COMPLETED = 'SEARCHING_FOR_ELEMENTS_COMPLETED';
 
 // Attributes on nodes that we know are unique to the node
 const uniqueAttributes = [
@@ -220,7 +226,6 @@ export function clearRecording () {
 export function getSavedActionFramework () {
   return async (dispatch) => {
     let framework = await settings.get(SAVED_FRAMEWORK);
-    console.log(framework);
     dispatch({type: SET_ACTION_FRAMEWORK, framework});
   };
 }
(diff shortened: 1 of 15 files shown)


dpgraham:
PR is ready now.

'sendKeys' doesn't do anything right now, that will be for the next PR

app/renderer/actions/Inspector.js
```diff
@@ -259,3 +266,47 @@ export function setSessionDetails (sessionDetails) {
...
+export function setLocatorTestStrategy (locatorTestStrategy) {
+  return (dispatch) => {
+    dispatch({type: SET_LOCATOR_TEST_STRATEGY, locatorTestStrategy});
+  };
+}
+
+export function searchForElement (locatorTestStrategy, locatorTestValue) {
+  return async (dispatch) => {
+    dispatch({type: SEARCHING_FOR_ELEMENTS});
+    const {result:elements} = await callClientMethod('elements', [locatorTestStrategy, locatorTestValue]);
```
jlipps: `{result: elements}`

app/renderer/components/Inspector/LocatorTestModal.js
```diff
@@ -0,0 +1,104 @@
...
+      ['-ios predicate string', 'Predicate String (iOS)'],
+      ['-ios class chain', 'Class Chain (iOS)'],
+    ];
+
+    return <Modal visible={isLocatorTestModalVisible} 
+      okText={locatedElements ? 'Done' : 'Search'}
+      cancelText='Cancel'
+      title='Search for element'
+      confirmLoading={isSearchingForElements}
+      onOk={(() => this.onSubmit()).bind(this)}
```
jlipps: curious why we need `bind` here? i thought `=>` preserved lexical scope?

app/renderer/components/Inspector/LocatorTestModal.js
```diff
@@ -0,0 +1,104 @@
...
+      onCancel={(() => this.onCancel()).bind(this)}>
+        {!locatedElements && <Row>
```
jlipps: we've got a few different sections here, each annotated with `locatedElements` or `!locatedElements`. it's fine but the pattern suggests maybe these are actually two different components?

app/renderer/lib/client-frameworks/ruby.js
```diff
@@ -55,6 +55,10 @@ driver.quit`;
   codeFor_back () {
     return `driver.back`;
   }
+
+  codeFor_clickElement () {
```
jlipps: yeah this is ok for now, but probably we don't want to need to handle `clickElement` when we already handle `click`

app/renderer/components/Inspector/LocatorTestModal.js
dpgraham: The method handlers bind `this` to `event`

jlipps: hmm, still seems wonky. would be nice if we could do one of these other things:

```js
onOk={this.onSubmit}
```

```js
onOk={() => this.onSubmit()}
```

```js
onOk={this.onSubmit.bind(this)} // if binding is necessary
```

jlipps:
also, do you know why the CI failed?

dpgraham:
It always fails due to code-signing.

app/renderer/lib/client-frameworks/ruby.js
dpgraham: This will be gone in a future PR.

dpgraham:
Fixes made.

The next PR will have the recorder updates. I just wanted to split them into two PR's because it would be pretty tough to review them all at once.

Ysoko:
Any update as to when this will be publicly available? Updated to Appium Desktop 1.1.1 today and still do not see this feature, very interested in being able to test XPATH's again.

PR #248, opened 2017-07-13
Make selected element panel scrollable

dpgraham:
* ANTD css is finicky so had to do a partial JS solution
* Sets the height of the selected element panel to the height of the window minus the distance of the panel to the top of the window minus a 10px buffer to account for the bottom padding
* Set the overflow to auto so that when it overflows it shows a vertical scrollbar
* Recalculates the height every time the SelectedElement component mounts and every time window size changes (debounced 300ms for performance reasons)


dpgraham:
@jlipps Have you had a chance to review this?

PR #249, opened 2017-07-14
Refactor recorder and main appium thread

dpgraham:
(review https://github.com/appium/appium-desktop/pull/239 first)

* Moved Appium server logic into a new method called AppiumMethodHandler which:
  * fetches individual elements
  * fetches multiple elements
  * caches element ID and remembers the selector, strategy used to find that element
  * keeps count of the fetches so that a variable name can be reported back to the element finder (e.g.: el1, el2, el3, ...)
* Updated recorder code so that it handles arrays and elements
* Separates 'find' and actions into two lines


app/renderer/components/Inspector/shared.js
```diff
@@ -13,4 +16,32 @@ export function parseCoordinates (element) {
...
+    return true;
+  }
+  const doc = new DOMParser().parseFromString(sourceXML);
+  return xpath.select(`//*[@${attrName}="${attrValue}"]`, doc).length < 2;
+}
+
+// Map of the optimal strategies. 
+const STRATEGY_MAPPINGS = [
+  ['name', 'accessibility id'],
+  ["content-desc", 'accessibility id'],
```
jlipps: why a different type of string quote for `content-desc`

app/renderer/actions/Inspector.js
```diff
@@ -142,23 +167,28 @@ export function applyClientMethod (params) {
...
-        // the element we are going to interact with. in the
-        // future we'll want to adjust this to use the locator
-        // strategy we recommend to the user, not just xpath
-        if (params.xpath) {
-          // also make sure we only do this optionally in case we're calling
-          // a global driver method that isn't operating on an element
-          recordAction('findElement', ['xpath', params.xpath])(dispatch);
+        // Add 'findAndAssign' line of code. Don't do it for arrays though. Arrays already have 'find' expression
+        if (strategy && selector && !variableIndex && variableIndex !== 0) {
+          findAndAssign(strategy, selector, variableName, false)(dispatch, getState);
```
jlipps: why pull `findAndAssign` out into its own action? it's only called here, right? unless you just wanted to make the code more modular/readable

dpgraham: It's called twice. In `searchForElement` and `applyClientMethod`

jlipps: oh ok i missed that

dpgraham:
PR fixes pushed

PR #260, opened 2017-07-28
Can swipe and tap on screenshot

dpgraham:
* Currently can only select items on a screenshot
* Now can either select, swipe or tap
* Tap does a tap touchaction on a absolute coordinates
* Swipe lets you select one coordinate, select a second coordinate and then it does a touch action like press -> moveTo -> release
* Shows the first coordinate as an SVG circle
* When both coordinates are selected, shows a rounded SVG line
* Updated recorder to handle these two actions

(https://drive.google.com/file/d/0ByQg0qBtDmn1N0hzUmZ3RXlYUXM/view?usp=sharing)

### DIFF
--- a/app/main/appium-method-handler.js
+++ b/app/main/appium-method-handler.js
@@ -82,10 +82,12 @@ export default class AppiumMethodHandler {
   async executeMethod (methodName, args = []) {
     let res = {};
 
-    // Specially handle the tap method (TODO: Add swipe here too)
+    // Specially handle the tap method
     if (methodName === 'tap') {
       res = await (new wd.TouchAction(this.driver)).tap({x: args[0], y: args[1]}).perform();
-      // res = await (new wd.TouchAction(driver)).press({el: el}).release();
+    } if (methodName === 'swipe') {
+      const [startX, startY, endX, endY]  = args;
+      res = await (new wd.TouchAction(this.driver)).press({x: startX, y: startY}).moveTo({x: endX, y: endY}).release().perform();
     } else if (methodName !== 'source' && methodName !== 'screenshot') {
       res = await this.driver[methodName].apply(this.driver, args);
     }
--- a/app/main/appium-method-handler.js
+++ b/app/main/appium-method-handler.js
@@ -82,12 +82,18 @@ export default class AppiumMethodHandler {
   async executeMethod (methodName, args = []) {
     let res = {};
 
-    // Specially handle the tap method
+    // Specially handle the tap and swipe method
     if (methodName === 'tap') {
-      res = await (new wd.TouchAction(this.driver)).tap({x: args[0], y: args[1]}).perform();
+      res = await (new wd.TouchAction(this.driver))
+        .tap({x: args[0], y: args[1]})
+        .perform();
     } else if (methodName === 'swipe') {
       const [startX, startY, endX, endY]  = args;
-      res = await (new wd.TouchAction(this.driver)).press({x: startX, y: startY}).moveTo({x: endX, y: endY}).release().perform();
+      res = await (new wd.TouchAction(this.driver))
+        .press({x: startX, y: startY})
+        .moveTo({x: endX, y: endY})
+        .release()
+        .perform();
     } else if (methodName !== 'source' && methodName !== 'screenshot') {
       res = await this.driver[methodName].apply(this.driver, args);
     }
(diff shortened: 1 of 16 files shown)


app/main/appium-method-handler.js
```diff
@@ -80,7 +81,14 @@ export default class AppiumMethodHandler {
 
   async executeMethod (methodName, args = []) {
     let res = {};
-    if (methodName !== 'source' && methodName !== 'screenshot') {
+
+    // Specially handle the tap method
```
jlipps: aren't we specially handling the tap _and_ swipe methods here?

app/main/appium-method-handler.js
```diff
@@ -80,7 +81,14 @@ export default class AppiumMethodHandler {
...
+    if (methodName === 'tap') {
+      res = await (new wd.TouchAction(this.driver)).tap({x: args[0], y: args[1]}).perform();
+    } else if (methodName === 'swipe') {
+      const [startX, startY, endX, endY]  = args;
+      res = await (new wd.TouchAction(this.driver)).press({x: startX, y: startY}).moveTo({x: endX, y: endY}).release().perform();
```
jlipps: i think this would look cleaner separating out the individual commands on separate lines:

```js
res = await (new wd.TouchAction(this.driver))
  .press({x: startX, y: starty})
  .moveTo({x: endX, y: endY})
  .release()
  .perform();
```

app/renderer/components/Inspector/Screenshot.js
```diff
@@ -35,15 +36,61 @@ export default class Screenshot extends Component {
...
+        x: Math.round(x),
+        y: Math.round(y),
+      });
     }
   }
 
+  handleMouseOut () {
+    this.setState({
+      ...this.state,
+      x: undefined,
```
jlipps: i don't like setting things to undefined, can we do `null` instead?

app/renderer/components/Inspector/Screenshot.js
```diff
@@ -89,20 +137,48 @@ export default class Screenshot extends Component {
     // If we're tapping or swiping, show the 'touch' cursor style
```
jlipps: the comment here doesn't seem to match what we're doing

app/renderer/components/Inspector/Screenshot.js
```diff
@@ -89,20 +137,48 @@ export default class Screenshot extends Component {
...
+        <div className={styles.screenshotActionsPanel}>
+          <Actions {...this.props} />
+        </div>
+        <div ref={(containerEl) => { this.containerEl = containerEl; }}
+          style={screenshotStyle} 
+          onClick={this.handleScreenshotClick.bind(this)}
+          onMouseMove={this.handleMouseMove.bind(this)}
+          onMouseOut={this.handleMouseOut.bind(this)}
+          className={styles.screenshotBox}>
+          {x !== undefined ? <div className={styles.coordinatesContainer}>
```
jlipps: this check could then be a more natural null check

app/renderer/components/Inspector/Screenshot.js
```diff
@@ -89,20 +137,48 @@ export default class Screenshot extends Component {
...
     const screenshotStyle = {};
     if (screenshotInteractionMode === 'tap' || screenshotInteractionMode === 'swipe') {
-      screenshotStyle.cursor = 'crosshair'; // TODO: Change this to touch
+      screenshotStyle.cursor = 'crosshair';
     }
 
     recursive(source);
 
     // Show the screenshot and highlighter rects. Show loading indicator if a method call is in progress.
     return <Spin size='large' spinning={!!methodCallInProgress}>
```
jlipps: there is now enough HTML code here I wonder if we should factor out some smaller components. up to you.

app/renderer/lib/client-frameworks/java.js
```diff
@@ -97,6 +97,14 @@ ${this.indent(code, 4)}
   codeFor_back () {
     return `driver.navigate().back();`;
   }
+
+  codeFor_tap (varNameIgnore, varIndexIgnore, x, y) {
+    return `(new TouchAction(driver)).tap(${x}, ${y}).perform()`;
+  }
+
+  codeFor_swipe (varNameIgnore, varIndexIgnore, x1, y1, x2, y2) {
+    return `(new TouchAction(driver)).press({x: ${x1}, y: ${y1}}).moveTo({x: ${x2}: y: ${y2}}).release().perform()`;
```
jlipps: i think we should also split this onto multiple lines and indent for the generated code as well.

dpgraham:
Fixes made

PR #263, opened 2017-08-03
Limit logs and write logs to file that can be opened by user

dpgraham:
* Limit logs to 1000
  * When logs get above 1000, the app slows down/crashes. Limit the number of rendered logs to 1000
* Log writing
  * Main thread writes all logs to file
  * Main thread sends the filepath of the logs to the browser window
  * Added 'GetRawLogsButton' to server monitor component
  * When button is clicked, opens the file externally
  * Delete logfile when browser closed
  * If logfile not present, show an alert to indicate that logs can't be accessed

### DIFF
--- a/app/main/appium.js
+++ b/app/main/appium.js
@@ -5,6 +5,7 @@ import { main as appiumServer } from 'appium';
 import { getDefaultArgs, getParser } from 'appium/build/lib/parser';
 import path from 'path';
 import wd from 'wd';
+import { fs, tempDir } from 'appium-support';
 import settings from '../settings';
 import autoUpdaterController from './auto-updater';
 import AppiumMethodHandler from './appium-method-handler';
@@ -20,9 +21,18 @@ var batchedLogs = [];
 let sessionDrivers = {};
 
 let appiumHandlers = {};
+let logFile;
 
 // Delete saved server args, don't start until a server has been started
 settings.deleteSync('SERVER_ARGS');
+
+async function deleteLogfile () {
+  if (logFile) {
+    try {
+      await fs.rimraf(logFile);
+    } catch (ign) { }
+  }
+}
 
 /**
  * Kill session associated with session browser window
@@ -46,6 +56,14 @@ async function killSession (sessionWinID) {
 
 function connectStartServer (win) {
   ipcMain.on('start-server', async (event, args) => {
+    // log the server logs to a file
+    try {
+      const dir = await tempDir.openDir();
+      logFile = path.resolve(dir, 'appium-server-logs.txt');
+      win.webContents.send('path-to-logs', logFile);
+      win.on('close', deleteLogfile);
+    } catch (ign) { }
+
     // clean up args object for appium log purposes (so it doesn't show in
     // non-default args list
     if (args.defaultCapabilities &&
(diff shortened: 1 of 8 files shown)


app/main/appium.js
```diff
@@ -60,9 +78,12 @@ function connectStartServer (win) {
     args.throwInsteadOfExit = true;
 
     // set up our log watcher
-    logWatcher = setInterval(() => {
+    logWatcher = setInterval(async () => {
       if (batchedLogs.length) {
-        win.webContents.send('appium-log-line', batchedLogs);
+        try {
+          await fs.writeFile(logFile, batchedLogs.map((log) => `[${log.level}] ${log.msg}`).join('\n'), {'flag': 'a'});
```
jlipps: don't need quotes around `flag`

jlipps: this would look better on multiple lines. one line for the map, the other for the write.

app/renderer/components/ServerMonitor/ServerMonitor.css
```diff
@@ -64,6 +64,16 @@
   border-color: #a0a0a0 !important;
 }
 
+.getRawLogsButton {
```
jlipps: could we share some of this code with `.stopButton`?

dpgraham:
Fixed the branch so that you can see the slicing in this commit: https://github.com/appium/appium-desktop/pull/263/commits/f0c997d232c0b68b1c6ea06f5d6c41f0f2b75cbc

dpgraham:
Fixes pushed

app/renderer/reducers/ServerMonitor.js
```diff
@@ -37,13 +40,21 @@ export default function serverMonitor (state = initialState, action) {
         sessionId: action.sessionUUID,
       };
     case LOGS_RECEIVED:
+      // TODO: We should dump logs to a txt file that can be exported
```
jlipps: this is no longer a TODO, right?

app/renderer/reducers/ServerMonitor.js
```diff
@@ -37,13 +40,21 @@ export default function serverMonitor (state = initialState, action) {
...
+      var logLines = state.logLines.concat(action.logs.map((l) => {
```
jlipps: `let`

dpgraham: That's an eslint rule. Apparently you can't declare using `let` inside of switch statements.

I'll just move the declaration outside of it instead.

imurchie: Yeah, case statements aren't blocks, so the variables are hoisted to the top of the switch statement, but uninitialized until the case it executed. This isn't so good.

The best solution would be to wrap the case stuff into a block.
```js
case LOGS_RECEIVED: {
  // TODO: We should dump logs to a txt file that can be exported
  let logLines = state.logLines.concat(action.logs.map((l) => {
    // attach a timestamp to each log line here when it comes in
    l.timestamp = moment().format('YYYY-MM-DD hh:mm:ss');
    return l;
  }));

  // Don't log more than MAX_LOG_LINES
  if (logLines.length > MAX_LOG_LINES) {
    logLines = logLines.slice(logLines.length - MAX_LOG_LINES);
  }

  return {
    ...state,
    logLines,
    serverStatus: STATUS_RUNNING
  };
}
```

PR #267, opened 2017-08-04
Feature: Edit the desired capabilities JSON string

dpgraham:
* Added another mode in the JSON representation that allows user to input JSON string
* Shows a textarea with buttons to cancel or save
* Shows errors if the JSON is invalid
* When click the editor button, translates the caps array to editable JSON
* When JSON is saved, translate it back to the caps array format
* Don't allow saving if an error occurs, show error message
* Disable the dcap forms when raw JSON is being edited to avoid conflicts

https://drive.google.com/a/saucelabs.com/file/d/0ByQg0qBtDmn1b3hfLUk5M0NoMm8/view?usp=sharing

### DIFF
--- a/app/renderer/actions/Session.js
+++ b/app/renderer/actions/Session.js
@@ -3,7 +3,7 @@ import settings from '../../settings';
 import { v4 as UUID } from 'uuid';
 import { push } from 'react-router-redux';
 import { notification } from 'antd';
-import { debounce } from 'lodash';
+import { debounce, toPairs } from 'lodash';
 import { setSessionDetails } from './Inspector';
 
 export const NEW_SESSION_REQUESTED = 'NEW_SESSION_REQUESTED';
@@ -45,6 +45,7 @@ export const ENABLE_DESIRED_CAPS_EDITOR = 'ENABLE_DESIRED_CAPS_EDITOR';
 export const ABORT_DESIRED_CAPS_EDITOR = 'ABORT_DESIRED_CAPS_EDITOR';
 export const SAVE_RAW_DESIRED_CAPS = 'SAVE_RAW_DESIRED_CAPS';
 export const SET_RAW_DESIRED_CAPS = 'SET_RAW_DESIRED_CAPS';
+export const SHOW_DESIRED_CAPS_JSON_ERROR = 'SHOW_DESIRED_CAPS_JSON_ERROR';
 
 export const ServerTypes = {
   local: 'local',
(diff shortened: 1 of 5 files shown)


app/renderer/reducers/Session.js
```diff
@@ -206,6 +219,79 @@ export default function session (state = INITIAL_STATE, action) {
...
+        rawDesiredCaps: formatJSON.plain(rawCaps),
+        isValidCapsJson: true,
+        isValidatingCapsJson: false, // Don't start validating JSON until the user has attempted to save the JSON
+      };
+
+    case ABORT_DESIRED_CAPS_EDITOR:
+      return {
+        ...state,
+        isEditingDesiredCaps: false,
+        rawDesiredCaps: undefined,
```
jlipps: can we set to `null` rather than `undefined`?

app/renderer/reducers/Session.js
```diff
@@ -206,6 +219,79 @@ export default function session (state = INITIAL_STATE, action) {
...
+      };
+
+    case SAVE_RAW_DESIRED_CAPS:
+      const {rawDesiredCaps} = state;
+      try {
+        const newCaps = JSON.parse(rawDesiredCaps);
+
+        // Transform the current caps array to an object
+        let {caps:capsArray} = state;
```
jlipps: `let {caps: capsArray} = state;`

app/renderer/reducers/Session.js
```diff
@@ -206,6 +219,79 @@ export default function session (state = INITIAL_STATE, action) {
...
+        let caps = {};
+        for (let {type, name, value} of capsArray) {
+          caps[name] = {type, value};
+        }
+
+        // Translate the caps JSON to array format
+        let newCapsArray = toPairs(newCaps).map(([name, value]) => ({
+          type: (caps[name] && caps[name].type === 'file' && typeof(value) === 'string') ? 'file' : getCapType(value),  // If we already have that cap and it's file type, keep the type the same
```
jlipps: this line is really long, it's fine to just make this a function rather than a lambda expression

jlipps: technically `typeof` is an operator, not a function, and i think we should standardize on not using parens. `typeof value`. (did this somewhere above too)

app/renderer/reducers/Session.js
```diff
@@ -206,6 +219,79 @@ export default function session (state = INITIAL_STATE, action) {
...
+          name,
+          value,
+        }));
+
+        return {
+          ...state,
+          isEditingDesiredCaps: false,
+          caps: newCapsArray,
+        };
+      } catch (e) {
```
jlipps: should we actually be doing json parsing and error handling here in the reducer? rather than say the action?

dpgraham: Yeah probably that would be better in the actions file.

dpgraham:
Changes pushed

app/renderer/actions/Session.js
```diff
@@ -419,3 +426,69 @@ export function getRunningSessions () {
...
+  };
+}
+
+export function abortDesiredCapsEditor () {
+  return (dispatch) => {
+    dispatch({type: ABORT_DESIRED_CAPS_EDITOR});
+  };
+}
+
+export function saveRawDesiredCaps () {
```
jlipps: yes, this is a much better place for all this!

PR #272, opened 2017-08-09
[Summary] Add ‘Clean Logs’ Button

Azuer:
Add 'Clean Logs' Button so than we can clean logs when appium server is running

### DIFF
--- a/app/renderer/components/ServerMonitor/ServerMonitor.js
+++ b/app/renderer/components/ServerMonitor/ServerMonitor.js
@@ -71,9 +71,23 @@ class GetRawLogsButton extends Component {
   }
 }
 
+class DeleteButton extends Component {
+  static propTypes = {
+    serverStatus: PropTypes.string.isRequired,
+  }
+
+  render () {
+    const {serverStatus, clearLogs} = this.props;
+    return serverStatus === STATUS_STOPPED ? null : <Button className={styles.serverButton}
+                                                      icon="pause-circle-o" 
+                                                      onClick={clearLogs}>Delete</Button>;
+  }
+}
+
 export default class ServerMonitor extends Component {
 
   static propTypes = {
+    clearLogs: PropTypes.func.isRequired,
     stopServer: PropTypes.func.isRequired,
     closeMonitor: PropTypes.func.isRequired,
     serverStatus: PropTypes.string.isRequired,
@@ -154,6 +168,7 @@ export default class ServerMonitor extends Component {
           <div className={`${styles['button-container']}`}>
             <StartSessionButton {...this.props} />
             <StopButton {...this.props} />
+            <DeleteButton {...this.props} />
           </div>
         </div>
         <div className={termClass} ref={(c) => this._term = c}>


app/renderer/components/ServerMonitor/ServerMonitor.js
```diff
@@ -71,9 +71,27 @@ class GetRawLogsButton extends Component {
...
+
+  render () {
+    const {serverStatus, clearLogs} = this.props;
+    if (serverStatus !== STATUS_STOPPED) {
+      return <Button className={styles.serverButton}
+              icon="pause-circle-o" 
+              onClick={clearLogs}>CLean Logs</Button>;
+    } else {
+      return null;
+    }
```
dpgraham: I don't think the `else { return null; }` is necessary

(EDIT: I stand corrected. The render method breaks if it returns undefined)

app/renderer/components/ServerMonitor/ServerMonitor.js
```diff
@@ -154,6 +172,7 @@ export default class ServerMonitor extends Component {
           <div className={`${styles['button-container']}`}>
             <StartSessionButton {...this.props} />
             <StopButton {...this.props} />
+            <ClearLogsButton {...this.props} />
```
dpgraham: Do you mind posting a screenshot so we can see what this looks like?

app/renderer/components/ServerMonitor/ServerMonitor.js
```diff
@@ -71,9 +71,27 @@ class GetRawLogsButton extends Component {
...
+  static propTypes = {
+    serverStatus: PropTypes.string.isRequired,
+  }
+
+  render () {
+    const {serverStatus, clearLogs} = this.props;
+    if (serverStatus !== STATUS_STOPPED) {
+      return <Button className={styles.serverButton}
+              icon="pause-circle-o" 
+              onClick={clearLogs}>CLean Logs</Button>;
```
dpgraham: I think the 'delete' icon would be a better choice

dpgraham:
Thanks! One addition is that in the `renderer/actions/ServerMonitor.js` file, `clearLogs` will need to empty the contents of the raw logs:

Current
```javascript
export function clearLogs () {
  return (dispatch) => {
    dispatch({type: LOGS_CLEARED});
  };
}
```

Next
```javascript
import { fs } from 'appium-support';

/* ... */

export function clearLogs () {
    const logfilePath = getState().startServer.logfilePath;
    await fs.writeFile(logfilePath, ''); 
    dispatch({type: LOGS_CLEARED});
  };
}
```

app/renderer/components/ServerMonitor/ServerMonitor.js
Azuer: image

Azuer: image

dpgraham:
I completely forgot that you can't use `fs` on the renderer thread (because it's a browser context). @Azuer you can take that revision out and we'll merge your changes, and I'll just make another PR on top of it that does the raw logs deletion.

Azuer:
Ok, I got it.
I’ve already token it out and rebased changes.

From: Dan Graham
Reply-To: appium/appium-desktop
Date: Thursday, 17 August 2017 at 23:52
To: appium/appium-desktop
Cc: Azuer, Mention
Subject: Re: [appium/appium-desktop] [Summary] Add ‘Clean Logs’ Button (#272)


I completely forgot that you can't use fs on the renderer thread (because it's a browser context). @Azuer<https://github.com/azuer> you can take that revision out and we'll merge your changes, and I'll just make another PR on top of it that does the raw logs deletion.

app/renderer/components/ServerMonitor/ServerMonitor.js
```diff
@@ -71,9 +71,27 @@ class GetRawLogsButton extends Component {
...
 }
 
+class DeleteButton extends Component {
+  static propTypes = {
+    serverStatus: PropTypes.string.isRequired,
+  }
+
+  render () {
+    const {serverStatus, clearLogs} = this.props;
+    if (serverStatus !== STATUS_STOPPED) {
```
mykola-mokhnach:
```js
return serverStatus === STATUS_STOPPED ? null : <Button className={styles.serverButton}
                                                                                         icon="pause-circle-o" onClick={clearLogs}>Delete</Button>;
```

Azuer: Ok,I've changed as you suggested and rebased the commit

PR #280, opened 2017-08-25
add sauce connect proxy option to sauce server config

jlipps:
enables sending all http commands via sc proxy (helpful when users can't hit ondemand.saucelabs.com directly from their work machines due to firewall restrictions, but sc proxy has been approved already).

also fixed up some visual design of the server monitor that was bugging me.

### DIFF
--- a/app/renderer/components/Session/Session.js
+++ b/app/renderer/components/Session/Session.js
@@ -3,7 +3,7 @@ import React, { Component } from 'react';
 import NewSessionForm from './NewSessionForm';
 import SavedSessions from './SavedSessions';
 import AttachToSession from './AttachToSession';
-import { Tabs, Form, Input, Button, Spin, Card, Icon, Col, Checkbox, Select } from 'antd';
+import { Tabs, Form, Input, Button, Spin, Card, Icon, Col, Checkbox, Select, Row } from 'antd';
 import { ServerTypes } from '../../actions/Session';
 import SessionStyles from './Session.css';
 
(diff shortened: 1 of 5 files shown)


app/renderer/components/Session/Session.js
```diff
@@ -69,12 +69,33 @@ export default class Session extends Component {
...
+                    <FormItem>
+                      <Input id='sauceUsername' placeholder={process.env.SAUCE_USERNAME ? 'Using data found in $SAUCE_USERNAME' : 'your-username'} addonBefore="Sauce Username" value={server.sauce.username} onChange={(e) => setServerParam('username', e.target.value)} />
+                    </FormItem>
+                    <FormItem>
+                      <Input id='saucePassword' type='password' placeholder={process.env.SAUCE_ACCESS_KEY ? 'Using data found in $SAUCE_ACCESS_KEY' : 'your-access-key'} addonBefore="Sauce Access Key" value={server.sauce.accessKey} onChange={(e) => setServerParam('accessKey', e.target.value)} />
+                    </FormItem>
+                  </Col>
+                  <Col span={10}>
+                    <FormItem>
+                      <Checkbox checked={!!server.sauce.useSCProxy} onChange={(e) => setServerParam('useSCProxy', e.target.checked)}> Proxy through Sauce Connect's Selenium Relay</Checkbox>
```
mykola-mokhnach: it is just github parser, that does't like quotes?

dpgraham: Yeah that's not an error.

Issue #319, opened 2017-10-18
appium desktop: tap,sendkeys,clear buttons are disable for every element in my App


Shahunks:
## Appium or Appium Desktop? - Appium Desktop

You are reporting an issue at the Appium Desktop repository. Appium Desktop is a wrapper around Appium. If you are having trouble running tests, it is much more likely that the problem you are encountering is not a problem with Appium Desktop but with Appium. For that reason we require you to have tried your testcase with Appium CLI before reporting issues. Check the checkbox below to confirm that you have proven your issue does _not_ reproduce on Appium itself:

- [ yes] I have verified that my issue does not occur with Appium and should be investigated as an Appium Desktop issue

## The problem

The problem

appium desktop: tap,sendkeys,clear buttons are disable for every element in my App

Environment

Appium version (or git revision) that exhibits the issue:1.6.5
Last Appium version that did not exhibit the issue (if applicable):
Desktop OS/version used to run Appium: MAC OS
Node.js version (unless using Appium.app|exe):6.10.1
Mobile platform/version under test: 8.0.0
Real device or emulator/simulator: emulator
Appium CLI or Appium.app|exe: appium.app
## Environment

- I am running Appium Desktop version 1.7.1
![screen shot 2017-10-18 at 3 32 43 pm](https://user-images.githubusercontent.com/25055088/31715693-79c4762c-b422-11e7-924b-0cd4558f666b.png)

- I am on (pick one):
     Mac

dpgraham:
Can you open up the dev tools (F12) and look what happens in the console logs when you select an element?

Shahunks:
well, there is nothing that happens in dev tools when i select element and all the Attributes value is false.
![screen shot 2017-10-18 at 3 32 43 pm](https://user-images.githubusercontent.com/25055088/31732006-90f1be22-b454-11e7-8a37-a08b241e3e58.png)

dpgraham:
In dev tools, is there a tab called `Redux`? 

That shows the logs for the application.

Shahunks:
I don't see a Redux tab.
![screen shot 2017-10-18 at 10 48 43 pm](https://user-images.githubusercontent.com/25055088/31732566-91376b5a-b456-11e7-9174-4aad1d586881.png)

dpgraham:
Right, it's not available in prod.

Shahunks:
So what could be the issue, user name text box is greyed out

dpgraham:
[...]

PR #324, opened 2017-10-23
Reload source when selected element is stale

dpgraham:
* If an element is stale, when you select an element it will try to find it and won't find it in the source
* This PR, if it encounters a stale element, automatically reloads the source 

(addresses https://github.com/appium/appium-desktop/issues/319)

p.s.: @mykola-mokhnach @imurchie I know A.D. is out of your domain, but with JLipps away I need somebody to have a look at the PR's

### DIFF
--- a/app/renderer/actions/Inspector.js
+++ b/app/renderer/actions/Inspector.js
@@ -138,14 +138,14 @@ export function selectElement (path) {
     if (!elementId) {
       // If the element is stale, unselect and reload the source
       dispatch({type: UNSELECT_ELEMENT});
-      await applyClientMethod({methodName: 'source'})(dispatch);
-    } else {
-      // Set the elementId, variableName and variableType for the selected element 
-      // (check first that the selectedElementPath didn't change, to avoid race conditions)
-      if (getState().inspector.selectedElementPath === path) {
-        dispatch({type: SET_SELECTED_ELEMENT_ID, elementId, variableName, variableType});
-      }
-    }    
+      return await applyClientMethod({methodName: 'source'})(dispatch);
+    } 
+
+    // Set the elementId, variableName and variableType for the selected element 
+    // (check first that the selectedElementPath didn't change, to avoid race conditions)
+    if (getState().inspector.selectedElementPath === path) {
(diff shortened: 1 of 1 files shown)


app/renderer/actions/Inspector.js
```diff
@@ -130,17 +130,22 @@ export function selectElement (path) {
...
-    // Set the elementId, variableName and variableType for the selected element 
-    // (check first that the selectedElementPath didn't change, to avoid race conditions)
-    if (getState().inspector.selectedElementPath === path) {
-      dispatch({type: SET_SELECTED_ELEMENT_ID, elementId, variableName, variableType});
-    }
-    
+    if (!elementId) {
+      // If the element is stale, unselect and reload the source
+      dispatch({type: UNSELECT_ELEMENT});
+      await applyClientMethod({methodName: 'source'})(dispatch);
```
mykola-mokhnach: So we reset the selection for this case?

dpgraham: Yeah. Because the element is stale, we dispatch UNSELECT_ELEMENT so that there isn't a selected element anymore.

mykola-mokhnach: Isn't it possible to restore the selection after reload happens?

app/renderer/actions/Inspector.js
```diff
@@ -130,17 +130,22 @@ export function selectElement (path) {
...
       selector: optimalSelector,
     });
 
-    // Set the elementId, variableName and variableType for the selected element 
-    // (check first that the selectedElementPath didn't change, to avoid race conditions)
-    if (getState().inspector.selectedElementPath === path) {
-      dispatch({type: SET_SELECTED_ELEMENT_ID, elementId, variableName, variableType});
-    }
-    
+    if (!elementId) {
```
mykola-mokhnach: IMHO positive conditions are more readable. Or we can do a fast return here, like

```
if (!element) {
   ....
   return;
}
```

dpgraham: Agreed, yeah I changed it.

PR #325, opened 2017-10-23
Debounce calls to selectElement

dpgraham:
* Moved element finding logic into a debounced function so that if an element is selected shortly after another one we don't do repeated calls to findElement

### DIFF
--- a/app/renderer/actions/Inspector.js
+++ b/app/renderer/actions/Inspector.js
@@ -114,7 +114,7 @@ const findElement = _.debounce(async function (strategy, selector, dispatch, get
     selector,
   });
 
-    // If the element is stale, unselect and reload the source
+  // If the element is stale, unselect and reload the source
   if (!elementId) {
     dispatch({type: UNSELECT_ELEMENT});
     return await applyClientMethod({methodName: 'source'})(dispatch);
@@ -128,7 +128,7 @@ const findElement = _.debounce(async function (strategy, selector, dispatch, get
 }, 1000);
 
 export function selectElement (path) {
-  return (dispatch, getState) => {
+  return async (dispatch, getState) => {
     // Set the selected element in the source tree
     dispatch({type: SELECT_ELEMENT, path});
     const state = getState().inspector;
@@ -153,7 +153,7 @@ export function selectElement (path) {
     const [optimalStrategy, optimalSelector] = strategyMap.length > 0 ? strategyMap[strategyMap.length - 1] : ['xpath', selectedElementXPath];
 
     // Debounce find element so that if another element is selected shortly after, cancel the previous search
-    findElement(optimalStrategy, optimalSelector, dispatch, getState, path);
+    await findElement(optimalStrategy, optimalSelector, dispatch, getState, path);
   };
 }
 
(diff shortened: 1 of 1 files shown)


app/renderer/actions/Inspector.js
```diff
@@ -106,14 +106,36 @@ export function bindAppium () {
...
+  // Set the elementId, variableName and variableType for the selected element 
+  // (check first that the selectedElementPath didn't change, to avoid race conditions)
+  if (getState().inspector.selectedElementPath === path) {
+    dispatch({type: SET_SELECTED_ELEMENT_ID, elementId, variableName, variableType});
+  }
+}, 1000);
+
 export function selectElement (path) {
-  return async (dispatch, getState) => {
+  return (dispatch, getState) => {
```
mykola-mokhnach: I'd keep async

app/renderer/actions/Inspector.js
```diff
@@ -125,27 +147,13 @@ export function selectElement (path) {
...
-      return await applyClientMethod({methodName: 'source'})(dispatch);
-    } 
-
-    // Set the elementId, variableName and variableType for the selected element 
-    // (check first that the selectedElementPath didn't change, to avoid race conditions)
-    if (getState().inspector.selectedElementPath === path) {
-      dispatch({type: SET_SELECTED_ELEMENT_ID, elementId, variableName, variableType});
-    }
+    // Debounce find element so that if another element is selected shortly after, cancel the previous search
+    findElement(optimalStrategy, optimalSelector, dispatch, getState, path);
```
mykola-mokhnach: shouldn't we await for it?

app/renderer/actions/Inspector.js
```diff
@@ -106,14 +106,36 @@ export function bindAppium () {
...
 
+// A debounced function that calls findElement and gets info about the element
+const findElement = _.debounce(async function (strategy, selector, dispatch, getState, path) {
+  // Get the information about the element
+  let {elementId, variableName, variableType} = await callClientMethod({
+    strategy,
+    selector,
+  });
+
+    // If the element is stale, unselect and reload the source
```
mykola-mokhnach: extra space here

app/renderer/actions/Inspector.js
dpgraham: The 'debounced' function doesn't return a promise, it gets called 1000 ms after the most recent call to it.

app/renderer/actions/Inspector.js
dpgraham: We have a linting rule that if there's nothing that's awaited inside of a function, don't use async.

app/renderer/actions/Inspector.js
mykola-mokhnach: The documentation says 
> Subsequent calls to the debounced function return the result of the last func invocation.

https://lodash.com/docs/4.17.4#debounce

dpgraham: Okay that makes sense then that if ALL of the calls to debounce return the promise then we can use await. It would be problem if the subsequent calls to debounce terminated and didn't return anything.

dpgraham:
Updates made

PR #326, opened 2017-10-23
Don't allow to be run from outside Applications folder

dpgraham:
* Don't allow to be run from outside Applications folder 
    * Added 'WrongFolder' component that is shown if user is not in applications folder on Mac that shows a 'Move to' button that moves the application programatically
    * When user clicks 'Move to' button, send message to main thread which calls `app.moveToApplicationsFolder`
    * Skip the wrong folder warning if it's an E2E test or if it's in dev
* Upgraded to Electron 1.8.1 (note that 1.8 is in beta so we'll have to hold off on cherry-picking this to a release until it's out of beta)
* Replaced `electron-prebuilt` with `electron`
* Fixed broken NPM scripts that resulted because `babel-register` is no longer in `.bin`


mykola-mokhnach:
finally we got some time to have this feature 👍

app/renderer/index.js
```diff
@@ -3,17 +3,37 @@ import { render } from 'react-dom';
...
 import configureStore from './store/configureStore';
 import './styles/app.global.css';
 import './styles/github-gist-theme.global.css';
 
+const {app} = electron.remote;
+const isDev = process.env.NODE_ENV === 'development';
 const store = configureStore();
 const history = syncHistoryWithStore(hashHistory, store);
 
+function shouldShowWrongFolderComponent () {
```
mykola-mokhnach: should we consider operating systems different from Mac OS?

app/renderer/components/WrongFolder/WrongFolder.js
```diff
@@ -0,0 +1,20 @@
...
+export default class WrongFolder extends Component {
+  // This tells the main thread to move this to applications folder which will cause the app to close and restart
+  moveToApplicationsFolder () {
+    ipcRenderer.send('appium-move-to-applications-folder');
+  }
+
+  render () {
+    return <div className={WrongFolderStyles['wrong-folder']}>
+      <div>
+        <div>Appium Desktop should be run from the Applications folder</div>
```
mykola-mokhnach: Do we show the same message on Linux/Windows?

app/renderer/index.js
dpgraham: Yes. My first commit didn't account for that and threw a 'not a function error' on the Windows build (via AppVeyor). So I added the check that the function exists which accounts for non Mac.

app/renderer/components/WrongFolder/WrongFolder.js
dpgraham: No the function only applies to Mac.

app/renderer/index.js
mykola-mokhnach: > So I added the check that the function exists which accounts for non Mac.

Can you please show which line of this PR does this OS verification?

dpgraham: It doesn't verify the OS explicitly. It verifies that `isInApplicationsFolder` is defined. It's not defined in non Mac operating systems which I discovered through failed AppVeyor test from previous commits.

mykola-mokhnach: then it worth adding a comment

app/renderer/components/WrongFolder/WrongFolder.js
mykola-mokhnach: do we exit the app if this dialog is canceled?

dpgraham: Yeah. Do you think we should give them the option to proceed without moving the folder?

mykola-mokhnach: definitely not ;)

PR #330, opened 2017-10-24
Use multiple selector strategies if first one doesn't work

dpgraham:
* Don't refresh page if we couldn't find element, instead just try the next locator strategy
* If nothing is found, show a notification recommending refreshing instead of just refreshing
* The previous refreshing feature was a mistake, because if the element can never be found it will ALWAYS refresh

### DIFF
--- a/app/renderer/actions/Inspector.js
+++ b/app/renderer/actions/Inspector.js
@@ -107,24 +107,27 @@ export function bindAppium () {
 
 
 // A debounced function that calls findElement and gets info about the element
-const findElement = _.debounce(async function (strategy, selector, dispatch, getState, path) {
-  // Get the information about the element
-  let {elementId, variableName, variableType} = await callClientMethod({
-    strategy,
-    selector,
-  });
+const findElement = _.debounce(async function (strategyMap, dispatch, getState, path) {
+  for (let [strategy, selector] of strategyMap) {
+    // Get the information about the element
+    let {elementId, variableName, variableType} = await callClientMethod({
+      strategy,
+      selector,
+    });
 
-  // If the element is stale, unselect and reload the source
-  if (!elementId) {
-    dispatch({type: UNSELECT_ELEMENT});
-    return await applyClientMethod({methodName: 'source'})(dispatch);
-  } 
-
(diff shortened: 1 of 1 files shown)


app/renderer/actions/Inspector.js
```diff
@@ -150,10 +153,10 @@ export function selectElement (path) {
 
     // Find the optimal selection strategy. If none found, fall back to XPath.
     const strategyMap = _.toPairs(getLocators(selectedElementAttributes, sourceXML));
```
mykola-mokhnach: I wouldn't say it is a map, but rather a list of pairs/entries

app/renderer/actions/Inspector.js
```diff
@@ -107,24 +107,27 @@ export function bindAppium () {
...
 
-  // Set the elementId, variableName and variableType for the selected element 
-  // (check first that the selectedElementPath didn't change, to avoid race conditions)
-  if (getState().inspector.selectedElementPath === path) {
-    dispatch({type: SET_SELECTED_ELEMENT_ID, elementId, variableName, variableType});
+    // Set the elementId, variableName and variableType for the selected element 
+    // (check first that the selectedElementPath didn't change, to avoid race conditions)
+    if (elementId && getState().inspector.selectedElementPath === path) {
+      return dispatch({type: SET_SELECTED_ELEMENT_ID, elementId, variableName, variableType});
+    }
```
mykola-mokhnach: what if none of the strategies matched?

app/renderer/actions/Inspector.js
```diff
@@ -107,24 +107,27 @@ export function bindAppium () {
...
   }
+
+  // If we couldn't find the element, show a notification
+  notification.error({
+    message: 'Error',
+    description: 'Could not find element. Try refreshing page',
```
mykola-mokhnach: I'd rather put Refresh button directly on this alert

mykola-mokhnach: the element, the page

sindhuiOS:
what the are the steps to perform iOS App Testing Using Appium

PR #344, opened 2017-11-15
Highlight elements that user searched for

dpgraham:
When the user opens the 'Search for Element' modal, and searches for elements, when they select one, highlight that element on the screenshots panel.

### DIFF
--- a/app/main/appium-method-handler.js
+++ b/app/main/appium-method-handler.js
@@ -58,7 +58,7 @@ export default class AppiumMethodHandler {
     return {variableName, variableType, strategy, selector, elements};
   }
 
-  async executeElementCommand (elementId, methodName, args = []) {
+  async executeElementCommand (elementId, methodName, args = [], skipScreenshot = false) {
     const cachedEl = this.elementCache[elementId];
 
     // Give the cached element a variable name (el1, el2, el3,...) the first time it's used
@@ -70,7 +70,10 @@ export default class AppiumMethodHandler {
     // Give the source/screenshot time to change before taking the screenshot
     await Bluebird.delay(500);
 
-    let sourceAndScreenshot = await this._getSourceAndScreenshot();
+    let sourceAndScreenshot;
+    if (!skipScreenshot) {
+      sourceAndScreenshot = await this._getSourceAndScreenshot();
+    }
 
     return {
       ...sourceAndScreenshot,
@@ -79,7 +82,7 @@ export default class AppiumMethodHandler {
     };
   }
 
-  async executeMethod (methodName, args = []) {
+  async executeMethod (methodName, args = [], skipScreenshot = false) {
     let res = {};
 
     // Specially handle the tap and swipe method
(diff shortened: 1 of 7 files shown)


app/main/appium-method-handler.js
```diff
@@ -70,7 +70,10 @@ export default class AppiumMethodHandler {
     // Give the source/screenshot time to change before taking the screenshot
     await Bluebird.delay(500);
 
-    let sourceAndScreenshot = await this._getSourceAndScreenshot();
+    let sourceAndScreenshot;
```
mykola-mokhnach: I assume an empty arrayshould be assigned to this variable since we apply ellipsis operator to it later

app/main/appium-method-handler.js
```diff
@@ -101,7 +104,11 @@ export default class AppiumMethodHandler {
...
```
mykola-mokhnach: can we avoid code duplication here?

app/main/appium.js
```diff
@@ -301,6 +301,7 @@ function connectClientMethodListener () {
       fetchArray = false, // Optional. Are we fetching an array of elements or just one?
       elementId, // Optional. Element being operated on 
       args = [], // Optional. Arguments passed to method
+      skipScreenshot = false, // Optional. Do we want the updated source and screenshot?
```
mykola-mokhnach: perhaps, we can select better name for this variable if it also affects source update?

app/renderer/actions/Inspector.js
```diff
@@ -377,11 +382,19 @@ export function findAndAssign (strategy, selector, variableName, isArray) {
   };
 }
 
-
-// TODO: Is this obsolete? I don't think we use this.
 export function setLocatorTestElement (elementId) {
```
mykola-mokhnach: shouldn't this function be also defined as async?

app/main/appium-method-handler.js
dpgraham: If the ellipsis operator has no affect on an object if it's `undefined` or `null`.

app/renderer/actions/Inspector.js
dpgraham: The function that is returned with is `async`.

app/main/appium.js
dpgraham: Good call

app/main/appium-method-handler.js
dpgraham: Yeah I'll make it so there's one main method.

dpgraham:
Changes pushed

app/main/appium-method-handler.js
```diff
@@ -58,20 +58,40 @@ export default class AppiumMethodHandler {
...
+    let cachedEl;
+    let res = {};
 
-    // Give the cached element a variable name (el1, el2, el3,...) the first time it's used
-    if (!cachedEl.variableName && cachedEl.variableType === 'string') {
-      cachedEl.variableName = `el${this.elVariableCounter++}`;
+    if (elementId) {
+      // Give the cached element a variable name (el1, el2, el3,...) the first time it's used
+      cachedEl = this.elementCache[elementId];
+      if (!cachedEl.variableName && cachedEl.variableType === 'string') {
```
mykola-mokhnach: you can use `_.isString`

mykola-mokhnach:
Any chance to fix Travis run?

dpgraham:
Unfortunately there's a Travis problem that breaks the Electron build. I'm leaving it there in the hopes that eventually Travis gets fixed.

app/main/appium-method-handler.js
dpgraham: The `variableType` is a hard-coded string that identifies what type the element is. It's used by the RecordedActions component which displays recorded code and it uses the variableType to determine if it should be `el.text("String")` or `el.text(123)` etc...

PR #350, opened 2017-11-21
Remove element not found notification

dpgraham:
* Notification was kind of vague
* Replaced it with alert on the right panel that tells the user when they can't do the basic interactions with an element

### DIFF
--- a/app/renderer/components/Inspector/SelectedElement.js
+++ b/app/renderer/components/Inspector/SelectedElement.js
@@ -3,7 +3,7 @@ import ReactDOM from 'react-dom';
 import _ from 'lodash';
 import { getLocators } from './shared';
 import styles from './Inspector.css';
-import { Button, Row, Col, Input, Modal, Table, Alert } from 'antd';
+import { Button, Row, Col, Input, Modal, Table, Alert, Spin } from 'antd';
 
 const ButtonGroup = Button.Group;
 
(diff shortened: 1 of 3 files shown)


app/renderer/components/Inspector/SelectedElement.js
```diff
@@ -83,11 +83,18 @@ export default class SelectedElement extends Component {
       });
     }
 
+    const showInteractionsLoadingIndicators = !elementInteractionsNotAvailable && !elementId;
```
mykola-mokhnach: this condition is not very clear to me

app/renderer/components/Inspector/SelectedElement.js
```diff
@@ -83,11 +83,18 @@ export default class SelectedElement extends Component {
...
+
     return <div>
+      {elementInteractionsNotAvailable && <Row type="flex" gutter={10}> 
+        <Col>
+          <Alert type="info" message="Interactions not available for this element" showIcon />
```
mykola-mokhnach: perhaps `are not available`?

dpgraham: Yeah that's probably more clear.

app/renderer/components/Inspector/SelectedElement.js
dpgraham: Yeah it was unnecessary to have it outside of the JSX, fix pushed.

PR #418, opened 2018-02-15
Make test for no PrivateHeaders problem

dpgraham:
* Looks like when using electron-builder with Node 9 and building assets to the `release/` folder, PrivateHeaders does not get transferred
* Added a test that verifies the `PrivateHeaders` folder is there so that this does not occur in the future
* Also, edited the scripts in package.json so that all of the tests are run, node modules are deleted and installed and removed a redundancy

### DIFF
--- a/test/e2e/main-e2e.test.js
+++ b/test/e2e/main-e2e.test.js
@@ -4,6 +4,7 @@ import os from 'os';
 import path from 'path';
 import chai from 'chai';
 import chaiAsPromised from 'chai-as-promised';
+import dirCompare from 'dir-compare';
 import MainPage from './pages/main-page-object';
 
 const platform = os.platform();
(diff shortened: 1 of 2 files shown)


test/e2e/main-e2e.test.js
```diff
@@ -60,4 +61,13 @@ describe('application launch', function () {
     await client.pause(500);
     await client.getWindowCount().should.eventually.equal(initialWindowCount + 1);
   });
+
+  it('check that WebDriverAgent private headers folder (regression test for https://github.com/appium/appium-desktop/issues/417)', async function () {
+    // NOTE: This isn't really an "e2e" test, but the test has to be written here because the /release 
+    // folder needs to be built in order to run the test
+    if (platform === 'darwin') {
```
imurchie: I would rather the logic be to skip the test if it is on a system that is not Mac.
```js
if (platform === 'darwin') {
  return this.skip();
}
// now do the test
```

dpgraham:
@imurchie Maybe. Do you have any ideas on how to test this?

test/e2e/main-e2e.test.js
```diff
@@ -60,4 +61,15 @@ describe('application launch', function () {
     await client.pause(500);
     await client.getWindowCount().should.eventually.equal(initialWindowCount + 1);
   });
+
+  it('check that WebDriverAgent private headers folder (regression test for https://github.com/appium/appium-desktop/issues/417)', async function () {
```
jlipps: check that the folder does what

test/e2e/main-e2e.test.js
```diff
@@ -60,4 +61,15 @@ describe('application launch', function () {
...
+    // NOTE: This isn't really an "e2e" test, but the test has to be written here because the /release 
+    // folder needs to be built in order to run the test
+    if (platform === 'darwin') {
```
jlipps: isn't this logic backwards?

imurchie: Yes. @dpgraham just copied-and-pasted my copy-and-paste example.

test/e2e/main-e2e.test.js
```diff
@@ -60,4 +61,15 @@ describe('application launch', function () {
...
+      this.skip();
```
imurchie: Needs a return.

imurchie:
@dpgraham Maybe get a hash of the original WDA folder and compare it to the one produced?

dpgraham:
Fixes pushed

PR #432, opened 2018-02-28
Check NPM and Node versions before packaging

dpgraham:
* Add check-engine to package-ci script
* Require NPM 3 and Node 6 when running `package-ci` script
  * This is to prevent problems we had when using newer versions of NPM/Node
  * This package isn't published to NPM and these constraints don't apply to the version of Node used in the Electron main thread, thus having strict constraints isn't a big problem
  * This won't effect contributors, contributors can still use engines that are defined in `devEngines`. It's only the `package-ci` script that has the strict requirements
* Fix e2e test because it uses nested `node_modules`
* Standardize names of NPM scripts
* Fixed `npm run clean` to install `rimraf` before calling `rimraf` on `node_modules`
* Upgraded postcss to resolve peer dependency issue

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,11 +21,10 @@
   "homepage": "https://github.com/chentsulin/electron-react-boilerplate#readme",
   "main": "main.js",
   "scripts": {
-    "test": "npm run unit-test",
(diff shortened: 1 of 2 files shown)


jlipps:
looks good, except:

- conflict
- squash did not fully squash :-)

Issue #266, opened 2017-08-04
integrate Zeit's Hazel update server


jlipps:
Seems like this might fit our use case really well:

https://github.com/zeit/hazel

dpgraham:
Yeah when I get a chance I'll check it out. I'd like to resolve the updater issues.

dpgraham:
Resolved with v1.4.1

PR #435, opened 2018-03-02
Autoupdate overhaul

dpgraham:
* Added zeit server (https://hazel-server-pxufsrwofl.now.sh/)
* Switched from electron-userland/builder to native autoupdate
  * Need to use native to use zeit
  * Ripped out React components and just use native dialogs for autoupdate notifications
  * Process is:
    * Checks for update
    * If update available show notification saying so
    * If accepted, call `autoupdate.checkForUpdates()` and wait for download
    * When download complete ask user to restart app
* Checks for updates every hour and 10 seconds after first startup

(also related to https://github.com/appium/appium-desktop/issues/266)


app/main/auto-updater/auto-updater.js
```diff
@@ -0,0 +1,94 @@
...
+        // If they say yes, get the updates now
+        autoUpdater.checkForUpdates();
+      }
+    });
+  } else {
+    if (fromMenu) {
+      autoUpdater.emit('update-not-available');
+    } else {
+      // If no updates found check for updates every hour
+      await B.delay(60 * 1000);
```
mykola-mokhnach: is the interval in seconds or millis?

mykola-mokhnach: why don't we use setTimeout?

app/main/auto-updater/update-checker.js
```diff
@@ -0,0 +1,11 @@
+import request from 'request-promise';
+import { getFeedUrl } from './config';
+
+export async function getUpdate (currentVersion) {
+  const res = await request.get(getFeedUrl(currentVersion));
+  if (res) {
+    return JSON.parse(res);
+  } else {
```
mykola-mokhnach: else is redundant

app/main/auto-updater/update-checker.js
```diff
@@ -0,0 +1,11 @@
+import request from 'request-promise';
+import { getFeedUrl } from './config';
+
+export async function getUpdate (currentVersion) {
+  const res = await request.get(getFeedUrl(currentVersion));
```
mykola-mokhnach: I assume this may throw an exception in case of request failure

app/main/auto-updater/auto-updater.js
jlipps: ms; setTimeout does not return a promise and can't be awaited

app/main/appium.js
```diff
@@ -7,9 +7,9 @@ import path from 'path';
 import wd from 'wd';
 import { fs, tempDir } from 'appium-support';
 import settings from '../settings';
-import autoUpdaterController from './auto-updater';
 import AppiumMethodHandler from './appium-method-handler';
 import request from 'request-promise';
+import { checkNewUpdates } from './auto-updater/auto-updater';
```
jlipps: if you rename it to `index.js` you can just do `from './auto-updater'`;

app/main/auto-updater/auto-updater.js
```diff
@@ -0,0 +1,94 @@
...
+ * Check for new updates
+ */
+export async function checkNewUpdates (fromMenu) {
+  // autoupdate.checkForUpdates always downloads updates immediately 
+  // This method (getUpdate) let's us take a peek to see if there is an update 
+  // available before calling .checkForUpdates
+  const update = await getUpdate(app.getVersion());
+  if (update) {
+    let {name, notes, pub_date} = update;
+    pub_date = moment(pub_date).format('MMM Do YYYY, h:mma');
```
jlipps: even if `update` really has to have snake_case, we should name our own variables using camelCase

app/main/auto-updater/config.js
```diff
@@ -0,0 +1,8 @@
+const config = {
+  baseFeedUrl: `https://hazel-server-pxufsrwofl.now.sh`,
```
jlipps: is this a stable url? should we pay for the ability to run it on say update.appium.io?

test/integration/update-checker.integration-test.js
```diff
@@ -0,0 +1,31 @@
...
+
+chai.should();
+chai.use(chaiAsPromised);
+
+describe('appiumDriverExtender', function () {
+  let latestVersion;
+
+  before(async function () {
+    const latestReleaseUrl = `https://api.github.com/repos/appium/appium-desktop/releases/latest`;
+    const res = JSON.parse(await request.get(latestReleaseUrl, { headers: {'user-agent': 'node.js'} }));
```
jlipps: `{headers: {'user-agent': 'node.js'}}`

app/main/auto-updater/config.js
dpgraham: @jlipps It's hosted by https://zeit.co/now... not sure how stable that is. I ran this server by following the instructions on zeit/hazel git page.

app/main/auto-updater/update-checker.js
dpgraham: Yep. Good catch. It should return false on error.

app/main/auto-updater/auto-updater.js
dpgraham: I changed it to `60 * 60 * 1000` so that it checks for updates every hour and not every minute.

app/main/auto-updater/config.js
dpgraham: The URL itself is permanent and I'm pretty sure the server should be up-and-running permanently; if it ever were taken down, I think we would get notice first.

dpgraham:
@jlipps Is this good to merge?

appveyor.yml
```diff
@@ -1,5 +1,5 @@
 environment:
-  nodejs_version: "6"
+  nodejs_version: "7"
```
jlipps: why 7 and not 8?

dpgraham: 8 caused the issues with WebDriverAgent privateheaders

jlipps: ok

PR #446, opened 2018-03-06
Fix autoupdater on dev

dpgraham:
* When it's dev, make `checkNewUpdates` a no-op and skip over the autoupdate methods
* This prevents it from attempting to do an update when it's dev

(NOTE: The patch looks more complicated then it is, it just wraps the code in an if statement)

### DIFF
--- a/app/main/auto-updater/index.js
+++ b/app/main/auto-updater/index.js
@@ -43,7 +43,7 @@ if (!isDev) {
       }, (response) => {
         if (response === 0) {
           // If they say yes, get the updates now
-          autoUpdater.checkForUpdates();
+          getUpdates();
         }
       });
     } else {
@@ -88,9 +88,9 @@ if (!isDev) {
         `Must restart to apply the updates ` +
         `(note: it may take several minutes for Appium Desktop to install and restart)`,
     }, (response) => {
-      // If they say yes, get the updates now
+      // If they say yes, restart now
       if (response === 0) {
-        getUpdates();
+        autoUpdater.quitAndInstall();
       }
     });
   });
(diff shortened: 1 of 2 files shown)


app/main/auto-updater/index.js
```diff
@@ -7,88 +7,99 @@
...
+  autoUpdater.setFeedURL(getFeedUrl(app.getVersion()));
+
+  /** 
+   * Check for new updates
+   */
+  checkNewUpdates = async function (fromMenu) {
+    // autoupdate.checkForUpdates always downloads updates immediately 
+    // This method (getUpdate) let's us take a peek to see if there is an update 
+    // available before calling .checkForUpdates
+    const update = await getUpdate(app.getVersion());
```
jlipps: in my mind the naming is backwards here. `checkForUpdates` sounds like it's doing a peek, whereas `getUpdate` sounds like it's doing the actual download.

app/main/auto-updater/index.js
```diff
@@ -7,88 +7,99 @@
...
+    if (update) {
+      let {name, notes, pubDate:pubDate} = update;
```
jlipps: `pubDate:pubDate` shouldn't be necessary

app/main/auto-updater/index.js
dpgraham: Yeah, the native `checkForUpdates` does the actual download. I can change `getUpdate` to `checkUpdates`

app/main/auto-updater/index.js
dpgraham: Good catch, that should be `pub_date:pubDate`

PR #453, opened 2018-03-08
Windows fixes

dpgraham:
* Using `const {checkForUpdates:getUpdates} = autoUpdater` was causing problems because it appears to change the context and when `getUpdates` is called, `this` is undefined
* Updated some builder dependencies
* Fixed typos in integration test


dpgraham:
@jlipps Yeah I did it from Windows and it didn't log-in properly

PR #468, opened 2018-03-19
Remove Extra Promise Wrap

Bamieh:
_execute already returns a promise

### DIFF
--- a/app/main/appium-method-handler.js
+++ b/app/main/appium-method-handler.js
@@ -102,12 +102,12 @@ export default class AppiumMethodHandler {
     };
   }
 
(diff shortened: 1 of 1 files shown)


jlipps:
Thanks @Bamieh. You are correct, and your code is technically simpler. However as a project we have decided to _always_ include async and await keywords as a matter of style, to clearly indicate to the reader that a method is async. With your changes, it would not be immediately clear that we are dealing with promises/async. We would have to read the called code to figure that out. So yes, we add an extra layer of promiseness due to this style, for the sake of code readability.

PR #475, opened 2018-03-23
Clear the variable names failed

canfeit:
The _.toPairs return an array of object keyed-value pairs,the variableName delete operation should be
`delete elCache[1].variableName;`.
But I think,appropriate object's property values iteration should be 
`for (const elCache of Object.values(this.elementCache))`.


canfeit:
Have signed Contributor License Agreement.

PR #484, opened 2018-04-04
Update known-issues

dpgraham:
* Get rid of the auto-updater known issue
* Add the Windows long-path issue

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ from the downloaded DMG file to your own file system (the best place is the
 not supported, and will not work.
 
 ## Known Issues
-* Auto updater is [not working](https://github.com/appium/appium-desktop/issues/74) for versions < 1.4.1
+* Some Windows 10 Users experience a `PathTooLongException` when installing the EXE. The workaround for this is to update the setting on Windows to [enable long paths](https://superuser.com/questions/1119883/windows-10-enable-ntfs-long-paths-policy-option-missing)
 
 ## Usage Instructions
 


dpgraham:
This issue is due to not using `asar` compression and thus having long file names

Not sure what we can do about it though, because we need appium to be uncompressed to run shell commands.

jmazess:
Feature request: We would appreciate a feature enhancement to the auto-update module whereby we can disable it since we're operating in a managed environment. We would prefer to deploy updates in a controlled fashion. Plus, this would have the side benefit of disabling a module which seems to be flakey anyway.

dpgraham:
@jmazess Do you want user's to be able to disable it themselves?

jmazess:
@dpgraham Actually, since this is a managed enterprise environment, we'd prefer to control this administratively on the user's behalf. Whether the user can turn it on or off is not important as long as we can turn it off for them en masse in Application Packaging or some kind of configuration or policy enforcement. However, if the software is going to continue to be per-user only, then a per-user setting somewhere to switch off update checking (either a registry setting or a config file setting like Appium server) would be satisfactory.

While we're on the subject of enterprise management, it would be nice if we had the choice to install this software per-machine instead of per-user like all of the other Appium-related components because this would be easier for us to manage and to deploy updates. And it would work better in an SCCM Task Sequence.

dpgraham:
Okay, @jmazess can you open a couple of issues for this? Appium Desktop is an electron app, and I'm sure there's probably a configuration to enable that.

dpgraham:
@jmazess I think this would do the trick:

https://www.electron.build/configuration/nsis

I'll just make an extra build that disables updates and is configured to install `perMachine`.

I don't know when I'll be able to prioritize this though. Would you have any interest in helping with this?

jmazess:
Issues opened.

Those NSIS parameters look interesting/promising, but I thought the installer was using squirrel? Or is it merely a wrapper?

I couldn't make my customer wait for these changes, so I went ahead and packaged the per-user edition of Appium Desktop 1.4.1 for deployment. If these capabilities are added to a future version, we will revisit this the next time we have a request/opportunity to package Appium Desktop for our customer.

I would not be able to help implement these features, sorry. I already went beyond my regular job responsibilities by even joining this forum and proposing changes.

Thanks.

dpgraham:
@jmazess the idea would be to have two bundles available for download. The per machine bundle (NSIS) and the per user bundle (Squirrel).

PR #487, opened 2018-04-05
Remove unneeded unauthorized SSL fields

dpgraham:
These are already in the `Advanced Setting` collapsible


jlipps:
What is this collapsible you speak of?

dpgraham:
@jlipps It's an Advanced Settings collapsible inside of the server tab that has server configurations: `Use proxy` and `Allow Unauthorized Certificate`.

PR #489, opened 2018-04-06
Get Screen Size using WindowSize rather than coordinates of root element

simongranger:
The root element's size doesn't always reflect the screen size and this will cause coordinates and Highlighter Rectangles to mismatch what's on screen.

### DIFF
--- a/app/main/appium-method-handler.js
+++ b/app/main/appium-method-handler.js
@@ -110,7 +110,7 @@ export default class AppiumMethodHandler {
   }
   
   async _getSourceAndScreenshot () {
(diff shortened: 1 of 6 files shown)


dpgraham:
Thanks @simongranger .... do you have a second to sign the CLA?

app/main/appium-method-handler.js
```diff
@@ -119,7 +119,7 @@ export default class AppiumMethodHandler {
       }
       sourceError = e;
     }
-
+
```
dpgraham: Extra whitespace

dpgraham:
Looks good. Just one tiny issue.

PR #509, opened 2018-04-24
Add internal session keep-alive

dpgraham:
* In cases where users are using a cloud Appium provider, `newCommandTimeout` cannot be set to infinity (90s is the max)
* Workaround for this is to call `driver.sessionCapabilities` every 30 seconds to keep it alive
* If no new command has been called in the last 5 minutes though, pop up a modal asking the user if they want to keep their session running
* When that modal comes up, they have one minute to keep the session running

* Also did some clean-up, made it so killing a session happens in one place

### DIFF
--- a/test/integration/appium-method-handler.integration-test.js
+++ b/test/integration/appium-method-handler.integration-test.js
@@ -103,4 +103,7 @@ describe('appiumDriverExtender', function () {
 
     });
   });
+  describe('', function () {
+    
+  });
 });
--- a/test/integration/appium-method-handler.integration-test.js
+++ b/test/integration/appium-method-handler.integration-test.js
@@ -103,7 +103,4 @@ describe('appiumDriverExtender', function () {
 
     });
   });
-  describe('', function () {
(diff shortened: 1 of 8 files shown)


test/integration/appium-method-handler.integration-test.js
```diff
@@ -103,4 +103,7 @@ describe('appiumDriverExtender', function () {
 
     });
   });
+  describe('', function () {
```
jlipps: probably don't need an empty test

matanmoda:
Hello Den! @dpgraham 
Is there and option to cancel or delay the Session timeout? This is causing the timeout prompt to show every minute, and kills it after another 30.. this is really difficult to use.. every time I work in some other window, the session is killed.

dpgraham:
@matanmoda Yeah, I'm going to change this to be 5 minutes and I'm going to make it so that it only runs on cloud sessions (SauceLabs, Headspin, BrowseStack, etc...).

PR #510, opened 2018-04-25
Add BrowserStack server tab

mohitmun:
Add BrowserStack as an option to give users access to more cloud providers supporting Appium

Screenshots:
image
image 1


app/main/appium.js
```diff
@@ -156,7 +156,7 @@ function connectCreateNewSessionWindow (win) {
 export function createNewSessionWindow (win) {
   // Create and open the Browser Window
   let sessionWin = new BrowserWindow({
-    width: 920,
+    width: 1080,
     minWidth: 920,
```
jlipps: if the new screen can't be minimized to 920 px wide without squashing the list of cloud providers, we should reset the min-width to be 1080 as well

mohitmun: Done

PR #517, opened 2018-05-02
Fix 'getCapsObject' problem

dpgraham:
* The getCapsObject was expecting `json_object` before but was sometimes getting `object` as a cap type because the type was determined via `typeof`
* Need to preserve `json_object` though because if you had saved a cap as `json_object` it wouldn't be compatible
* Now it just accepts both

Next step is to update Electron version, make a PR for the electron-builder issue, and do a release that uses a hacked version of electron-builder (for now)


app/renderer/components/Session/NewSessionForm.js
```diff
@@ -25,14 +25,17 @@ export default class NewSessionForm extends Component {
...
+
     switch (cap.type) {
       case 'text': return <Input disabled={isEditingDesiredCaps} id={`desiredCapabilityValue_${index}`} placeholder='Value' value={cap.value} onChange={(e) => setCapabilityParam(index, 'value', e.target.value)} size="large"/>;
       case 'boolean': return <Switch disabled={isEditingDesiredCaps} id={`desiredCapabilityValue_${index}`} checkedChildren={'true'} unCheckedChildren={'false'}
         placeholder='Value' checked={cap.value} onChange={(value) => setCapabilityParam(index, 'value', value)} />;
       case 'number': return <Input disabled={isEditingDesiredCaps} id={`desiredCapabilityValue_${index}`} placeholder='Value' value={cap.value}
         onChange={(e) => !isNaN(parseInt(e.target.value, 10)) ? setCapabilityParam(index, 'value', parseInt(e.target.value, 10)) : setCapabilityParam(index, 'value', undefined)} size="large"/>;
-      case 'object': return <Input disabled={isEditingDesiredCaps} id={`desiredCapabilityValue_${index}`} type='textarea' rows={4} placeholder='Value' value={cap.value}
-        onChange={(e) => setCapabilityParam(index, 'value', e.target.value)} size="large"/>;
+      case 'object': return objectRes;
```
jlipps: this could look like:

```js
case 'object':
case 'json_object':
    return objectRes;
```

app/renderer/components/Session/NewSessionForm.js
```diff
@@ -67,6 +70,9 @@ export default class NewSessionForm extends Component {
       case 'number':
         translatedValue = parseInt(translatedValue, 10) || 0;
         break;
+      case 'json_object':
```
jlipps: same here; no need to duplicate the case body

PR #573, opened 2018-06-18
Sauce demo mode

dpgraham:
Add a flag that, when set, only displays Sauce Labs cloud providers.

This is only going to be used internally at Sauce Labs, won't have any effect on the actual Appium Desktop.

### DIFF
--- a/app/renderer/components/Session/Session.js
+++ b/app/renderer/components/Session/Session.js
@@ -57,12 +57,12 @@ export default class Session extends Component {
             <TabPane tab={testObjectTabHead} key={ServerTypes.testobject}>
               <ServerTabTestobject {...this.props} />
             </TabPane>
-            <TabPane tab={headspinTabHead} key={ServerTypes.headspin}>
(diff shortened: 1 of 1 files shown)


jlipps:
not entirely sure what i think of this. it's one thing for a cloud provider to propose a new feature of utilizing their service. it's another for a cloud provider to propose a new feature that, if adopted, hides the existence of competing services. i get that it's only a flag that would ever be used by sauce people, but it still seems egregiously partisan. i think i'd prefer one of these two alternatives:

1) this PR gets updated to provide environment flags for _all_ the cloud services, so it doesn't look like a sauce-vs-others thing
2) sauce maintains a fork with this flag in it for their own purposes

imurchie:
or
3. other cloud providers do the work in their own PRs to do something similar, with the proviso herein that they would be accepted.

dpgraham:
@jlipps Yeah I can change it so that it's the same for all providers. I'll update that today or tomorrow.

PR #579, opened 2018-07-04
Unhandled exception handling

dpgraham:
* Use `electron-unhandled` which catches uncaught exceptions/promise-rejections and then shows a dialog with the message
*

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ addons:
     packages:
     - xvfb
 before_script:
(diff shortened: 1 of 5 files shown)


jlipps:
is this PR now obsolete?

dpgraham:
@jlipps it is yeah.... electorn-unhandled seems to always show a dialog when an exception occurred, regardless of whether or not it was caught.

PR #580, opened 2018-07-04
Show scroll buttons on expanded paths change

dpgraham:
* Checks that expandedPaths have changed when deciding if we should show the horizontal scroll buttons
* Also, remove electron-unhandled because it gets called even for caught exceptions/promises

### DIFF
--- a/app/main/appium.js
+++ b/app/main/appium.js
@@ -2,6 +2,7 @@
 
 import { ipcMain, BrowserWindow, Menu, app } from 'electron';
 import { main as appiumServer } from 'appium';
+import unhandled from 'electron-unhandled';
 import { getDefaultArgs, getParser } from 'appium/build/lib/parser';
 import path from 'path';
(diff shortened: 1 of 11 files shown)


app/main/appium.js
```diff
@@ -143,10 +143,16 @@ function connectCreateNewSessionWindow (win) {
...
+  ipcMain.on('appium-clear-logfile', async (event, {logfilePath}) => {
+    await fs.writeFile(logfilePath, '');
+  });
+}
+
 export function createNewSessionWindow (win) {
   // Create and open the Browser Window
   let sessionWin = new BrowserWindow({
-    width: 1080,
+    width: 300, // TODO: DO NOT COMMIT THIS!!!!!
```
jlipps: don't commit this :-)

jlipps:
the change described in the PR looks ok but it seems there are other unrelated changes here.

PR #581, opened 2018-07-04
Add cloud provider visibility menu

dpgraham:
(Replaces an older PR)

* Add a submenu in 'View' that lets you toggle which cloud providers are visible

screen shot 2018-07-04 at 4 31 31 pm

* Toggling the submenu items changes the state using `electron-settings` so that the user preferences are persisted across sessions
* Created a `shared` folder that has the logic for cloud providers, eventually I'd like to move all of the cloud provider logic there so that there's only one place where we configure it

(cc: @llaskin)

### DIFF
--- a/app/main/main.development.js
+++ b/app/main/main.development.js
@@ -87,11 +87,11 @@ app.on('ready', async () => {
   });
 
   if (process.platform === 'darwin') {
-    template = menuTemplates.mac(mainWindow);
+    template = await menuTemplates.mac(mainWindow);
     menu = Menu.buildFromTemplate(template);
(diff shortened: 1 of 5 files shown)


jlipps:
code looks good, but why are we getting and setting state outside of the redux flow?

dpgraham:
@jlipps because the menu items aren't part of the React UI I couldn't bind the state between the menu and Redux.

I could have added a Redux step, but it seems like it would have been a needless extra step because it would just be checking the state in `electron-settings` and then updating the Redux state only the one time.

PR #586, opened 2018-07-20
Add support for copying source XML to clipboard

caguilar187:
Adds support for copying the source xml to the system clipboard. 

Main use case using the search for element can be slow due to the page source being refreshed. This allows a user to copy the xml to any other xpath tester available to them. We have been using this quite a bit and find it faster to work with. The copy button is in the tooltip section of the inspector.

Lastly, if someone knows why my package.json went all crazy I'm happy to update prior to merging if this gets accepted

### DIFF
--- a/app/renderer/components/Inspector/Inspector.js
+++ b/app/renderer/components/Inspector/Inspector.js
@@ -48,7 +48,7 @@ export default class Inspector extends Component {
     const {screenshot, screenshotError, selectedElement = {},
       applyClientMethod, quitSession, isRecording, showRecord, startRecording,
       pauseRecording, showLocatorTestModal, screenshotInteractionMode, 
-      showKeepAlivePrompt, keepSessionAlive} = this.props;
+      showKeepAlivePrompt, keepSessionAlive, sourceXML} = this.props;
     const {path} = selectedElement;
 
     let main = <div className={InspectorStyles['inspector-main']}>
@@ -125,7 +125,7 @@ export default class Inspector extends Component {
            <Button id='searchForElement' icon="search" onClick={showLocatorTestModal}/>
         </Tooltip>
         <Tooltip title="Copy XML Source to Clipboard">
-           <Button id='btnSourceXML' icon="copy" onClick={() => clipboard.writeText(this.props.sourceXML)}/>
+           <Button id='btnSourceXML' icon="copy" onClick={() => clipboard.writeText(sourceXML)}/>
         </Tooltip>
         <Tooltip title="Quit Session & Close Inspector">
           <Button id='btnClose' icon='close' onClick={() => quitSession()}/>
--- a/app/renderer/components/Inspector/Inspector.js
+++ b/app/renderer/components/Inspector/Inspector.js
@@ -7,7 +7,7 @@ import Source from './Source';
 import SourceScrollButtons from './SourceScrollButtons';
 import InspectorStyles from './Inspector.css';
 import RecordedActions from './RecordedActions';
-const {clipboard} = require('electron')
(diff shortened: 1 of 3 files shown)


app/renderer/components/Inspector/Inspector.js
```diff
@@ -123,6 +129,9 @@ export default class Inspector extends Component {
         <Tooltip title="Search for element">
            <Button id='searchForElement' icon="search" onClick={showLocatorTestModal}/>
         </Tooltip>
+        <Tooltip title="Copy XML Source to Clipboard">
+           <Button id='btnSourceXML' icon="copy" onClick={() => this.copySourceXMLToClipboard()}/>
```
jlipps: since `copySourceXMLToClipboard` has effectively only one line of code in it and because it's used only once, might as well just call `copy` here

package.json
```diff
@@ -107,6 +107,7 @@
     "appium-support": "^2.12.0",
     "babel-core": "^6.26.0",
     "bluebird": "^3.5.1",
+    "copy-to-clipboard": "^3.0.8",
```
jlipps: we already have functionality that copies stuff to the clipboard. i don't remember where it is offhand (i think it's in the recorder). but i'd prefer not to install an additional package when i'm pretty sure we have a way to do it already.

dpgraham:
@caguilar187 Are you able to post a screenshot?

caguilar187:
@jlipps comments pushed

@dpgraham screen shot attached
screen shot 2018-07-23 at 3 26 10 pm

dpgraham:
@caguilar187 Just one last comment and we'll be good to go.

app/renderer/components/Inspector/Inspector.js
```diff
@@ -7,6 +7,7 @@ import Source from './Source';
 import SourceScrollButtons from './SourceScrollButtons';
 import InspectorStyles from './Inspector.css';
 import RecordedActions from './RecordedActions';
+const {clipboard} = require('electron')
```
dpgraham:
```javascript
import {clipboard} from 'electron';
```

dpgraham:
@jlipps You have any more comments?

app/renderer/components/Inspector/Inspector.js
```diff
@@ -123,6 +124,9 @@ export default class Inspector extends Component {
         <Tooltip title="Search for element">
            <Button id='searchForElement' icon="search" onClick={showLocatorTestModal}/>
         </Tooltip>
+        <Tooltip title="Copy XML Source to Clipboard">
+           <Button id='btnSourceXML' icon="copy" onClick={() => clipboard.writeText(this.props.sourceXML)}/>
```
jlipps: we typically pull props out of `this.props` at the beginning of the method. then we can just refer to it as `sourceXML` here

jlipps:
one last comment otherwise LGTM!

dpgraham:
@jlipps that's right, I forgot about that...

@caguilar187 At the top of the render method, add `sourceXML` to this:

```javascript
const {screenshot, screenshotError, selectedElement = {},
       applyClientMethod, quitSession, isRecording, showRecord, startRecording,	       applyClientMethod, quitSession, isRecording, showRecord, startRecording,
       pauseRecording, showLocatorTestModal, screenshotInteractionMode, 	       pauseRecording, showLocatorTestModal, screenshotInteractionMode, 
       showKeepAlivePrompt, keepSessionAlive} = this.props;	       showKeepAlivePrompt, keepSessionAlive} = this.props;
```

caguilar187:
@dpgraham yep addressed last night

PR #588, opened 2018-07-24
ANDROID_HOME set from in UI

dpgraham:
* Added a 'Config' window that is either opened from menu or from the start server page
* Added form that sets environment variables JAVA_HOME and ANDROID_HOME
* Shows defaults as whatever 'appium-desktop' finds inside of `process.env` originally
* When these are saved, saves it to appium-settings
* When app is loaded for first time, `process.env` is extended with these saved env vars
* When config is 'Saved' the user is asked to 'Restart Now' for changes to take effect

* Other

  * Added 'Stubs' for actions, reducers and components for cutting-and-pasting purposes later on
  * Moved browser window opener to a helper method

* This is how it looks, we'll probably want to change it to something more subtle
screen shot 2018-07-24 at 4 12 52 pm

screen shot 2018-07-24 at 4 19 54 pm

### DIFF
--- a/app/main/appium.js
+++ b/app/main/appium.js
@@ -1,6 +1,6 @@
 /* eslint-disable no-console */
 
-import { ipcMain, BrowserWindow, Menu, app } from 'electron';
+import { ipcMain, app } from 'electron';
 import { main as appiumServer } from 'appium';
 import { getDefaultArgs, getParser } from 'appium/build/lib/parser';
 import path from 'path';
@@ -11,9 +11,9 @@ import settings from '../settings';
 import AppiumMethodHandler from './appium-method-handler';
 import request from 'request-promise';
 import { checkNewUpdates } from './auto-updater';
+import { openBrowserWindow } from './helpers';
 
 const LOG_SEND_INTERVAL_MS = 250;
-const isDev = process.env.NODE_ENV === 'development';
 
 const defaultEnvironmentVariables = _.clone(process.env);
 
@@ -23,7 +23,6 @@ var batchedLogs = [];
 
 let appiumHandlers = {};
 let logFile;
-let configWin;
 
 // Delete saved server args, don't start until a server has been started
 settings.deleteSync('SERVER_ARGS');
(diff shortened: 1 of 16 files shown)


dpgraham:
@jlipps Did you have a chance to look at this?

Also, are there ENV variables other than ANDROID_HOME, or JAVA_HOME that should be configurable?

jlipps:
@dpgraham trying to take a look now but i'm getting conflicts pulling the branch. can you fix then i'll play with it?

dpgraham:
FIxed

jlipps:
pulled and reviewing today

jlipps:
when i pulled this and ran it for the first time, i got `ANDROID_HOME` not defined error on opening the config window

jlipps:
seems like the config window should default to a much smaller size

app/main/appium.js
```diff
@@ -374,6 +355,39 @@ function connectMoveToApplicationsFolder () {
...
+  });
+}
+
+function connectSaveEnv () {
+  ipcMain.on('appium-save-env', async (event, environmentVariables) => {
+    // Pluck unset values
+    const env = _.pickBy(environmentVariables, _.identity);
+
+    await settings.set('ENV', env);
+    setEnv();
```
jlipps: `await`

app/renderer/actions/Stub.js
```diff
@@ -0,0 +1,8 @@
+export const NAME_OF_ACTION_1 = 'NAME_OF_ACTION_1';
```
jlipps: is this file needed?

app/renderer/components/Config/Config.js
```diff
@@ -0,0 +1,59 @@
...
+    const {environmentVariables} = this.props;
+    ipcRenderer.send('appium-save-env', environmentVariables);
+    ipcRenderer.once('appium-save-env-done', () => {
+      const message = `Application must be restarted for changes to take effect`;
+      const dialogOptions = {type: 'info', buttons: ['Restart Now', 'Restart Later'], message};
+      dialog.showMessageBox(dialogOptions, (response) => {
+        if (response === 0) {
+          // If 'Restart Now' restart the application
+          app.relaunch();
+          app.exit();
```
jlipps: why do we both relaunch and exit?

app/renderer/components/Config/Config.js
```diff
@@ -0,0 +1,59 @@
...
+          getCurrentWindow().close();
+        }
+      });
+    });
+  }
+
+  render () {
+    const {setEnvironmentVariable, environmentVariables, defaultEnvironmentVariables} = this.props;
+
+    const ENV_VARIABLE_NAMES = [
```
jlipps: put these up top?

app/renderer/components/Stub.js
```diff
@@ -0,0 +1,12 @@
+import React, { Component } from 'react';
```
jlipps: this stub file looks unnecessary

app/renderer/reducers/Stub.js
```diff
@@ -0,0 +1,28 @@
+import {
```
jlipps: unused file

app/renderer/actions/Stub.js
dpgraham: These are 'Stub' files for future usage for copying and pasting to create new actions/reducers/components

app/renderer/components/Config/Config.js
dpgraham: Not sure why we have to do that but apparently it's required: https://github.com/electron/electron/issues/6579

app/renderer/actions/Stub.js
jlipps: oh i see so independently of this PR you want to have stub files for developer assistance in the future.

app/renderer/components/Config/Config.js
jlipps: ok, as long as it's the way we're supposed to do it!

jlipps:
hey @dpgraham i'm heading out on vacation, so from my perspective if you fix the issues mentioned here it's good to go

mrk-han:
Does this mean that Appium Desktop is no longer reading from zshconfig/bash_profiles?

dpgraham:
@mrk-han It works the same as before, the only difference is that if you opt to set an environment variable from the UI, it will use that value instead of what you have set in your bash profile.

swarnock89:
@dpgraham This may not be working as intended. When I don't have paths set in the configuration menu, my tests don't start up. When I copy the paths from my .bash_profile and enter them in the configuration menu, my tests run fine.

PR #590, opened 2018-07-30
Fix error on variable and params generation on RobotFramework Appium

NeoMorfeo:
There was a commit not uploaded to last PullRequest due to an error on local git.

This is only something related with the code generation, not with application functionality or stability.


NeoMorfeo:
@jlipps you are right, my fault, sorry. I have created another PR to solve it, with only the change i want to merge.

Thanks for your care.

PR #632, opened 2018-09-24
Azure Pipelines

dpgraham:
* Added Azure Pipelines CI so that we can do builds on Mac OSX again
* Also added back package-lock and yarn.lock so that packages are deterministic


jeremyepling:
I'm a product manager on Azure Pipelines. I noticed that your Linux job is timing out. The default is 60 minutes, but you can expand it to 360 minutes. https://docs.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=vsts&tabs=yaml#timeouts

PR #636, opened 2018-09-25
Make killSession synchronous

dpgraham:
* Move method handlers into appium-method-handler as static methods
* Make it killSession synchronous because there was a race condition where sessions were being killed as new one was being created and caused a crash
* This method can be synchronous because we don't use the result and it's fine to 'fire-and-forget'. There's nothing that needs to be waited for

### DIFF
--- a/app/main/appium-method-handler.js
+++ b/app/main/appium-method-handler.js
@@ -1,5 +1,7 @@
 import Bluebird from 'bluebird';
 import wd from 'wd';
+import log from 'electron-log';
+import _ from 'lodash';
 
 const isDevelopment = process.env.NODE_ENV === 'development';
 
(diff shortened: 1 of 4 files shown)


app/main/appium-method-handler.js
```diff
@@ -213,4 +215,40 @@ export default class AppiumMethodHandler {
...
+  static createSession (driver, sender, winId) {
+    const {appiumHandlers} = AppiumMethodHandler;
+    log.info(`Creating method handler for session with window id: ${winId}`);
+    const handler = AppiumMethodHandler.appiumHandlers[winId] = new AppiumMethodHandler(driver, sender);
+    log.info(`The following session window ID's have active sessions: '${JSON.stringify(_.keys(appiumHandlers))}'`);
+    return handler;
+  }
+
+  static killSession (winId, killedByUser) {
+    const {appiumHandlers} = AppiumMethodHandler;
```
jlipps: why recreate all the `getSessionHandler` logic in here?

dpgraham: There was a race condition where the appiumHandlers sessions were being prematurely deleted. So to debug, I moved all the logic into that one class and then added logging to it so I can see when objects in the key were being get/set/deleted.

dpgraham: nvm, i understand what you're talking about now

PR #658, opened 2018-10-05
Linting fixes

dpgraham:
* Linter wasn't being run in CI so many lint errors were uncaught
* Added the lint task to `npm test`
* Fixed linting problems... nothing too serious was going uncaught mostly just unused variables, semicolons, etc...


app/renderer/components/ServerMonitor/ServerMonitor.js
```diff
@@ -53,12 +49,11 @@ class StopButton extends Component {
   }
 }
 
-class StartSessionButton extends Component {
-  static propTypes = {
```
jlipps: does our linter not allow static fields?

PR #670, opened 2018-10-13
default custom server port to 4723 which is the value in placeholder

ianweng:
The default values (`hostname`: `127.0.0.1` and `port`: `4723` described in the placeholders) are not applied which throws a `Could not start session`.

### DIFF
--- a/app/renderer/actions/Session.js
+++ b/app/renderer/actions/Session.js
@@ -173,7 +173,7 @@ export function newSession (caps, attachSessId = null) {
         break;
       case ServerTypes.remote:
         host = session.server.remote.hostname;
-        port = session.server.remote.port;
+        port = session.server.remote.port || 4723;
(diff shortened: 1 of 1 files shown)


dpgraham:
Good idea. One thing that will need to be added to this is adding the `placeholder` attribute to the host and port inputs.

ianweng:
**Before my change**
Host: input: '', placeholder: '127.0.0.1' => 'Can not start session'

**My change**
Host: input: '', placeholder: '127.0.0.1' => if input is empty use '127.0.0.1' to start session

**Are you suggesting?** @dpgraham 
Host: input: '127.0.0.1', placeholder: '127.0.0.1'

dpgraham:
Ahhh, never mind. I just noticed the placeholders were already there and you're correcting a mistake. I'll merge it now.

PR #691, opened 2018-10-29
Fix rendering of Emoji strings

dpgraham:
* Emojis in the XML breaks the XML DOM parser
* Other unicode characters don't break the parser
* To fix this, added a script that replaces `&#01234` with `#01234` so that we can still see what the Emoji code is but it won't break the XML

### DIFF
--- a/app/renderer/actions/Inspector.js
+++ b/app/renderer/actions/Inspector.js
@@ -71,6 +71,11 @@ const uniqueAttributes = [
  */
 function xmlToJSON (source) {
   let xmlDoc;
+  
+  // Replace strings with Unicode format &#012345 with #012345
+  // The &# unicode format breaks the parser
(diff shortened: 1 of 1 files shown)


app/renderer/actions/Inspector.js
```diff
@@ -71,6 +71,11 @@ const uniqueAttributes = [
  */
 function xmlToJSON (source) {
   let xmlDoc;
+  
+  // Replace strings with Unicode format &#012345 with #012345
+  // The &# unicode format breaks the parser
+  source = source.replace(/&#([0-9]*)/g, "#$1");
```
mykola-mokhnach: this regexp is going to replace all occurrences of `&#`. Shall we replace `[0-9]*` with `[0-9]{4,}`?

PR #752, opened 2018-12-03
Change from Webpack 1 to Parcel

dpgraham:
To get Appium 1.10 working, I needed to get babel 7 working which meant I needed to migrate from Webpack 1 to Webpack 4. I made the executive decision that since Weback 1 to 4 has a complicated migration path, we'd do just as well migrating to ParcelJS.

It uses ParcelJS now to build the main thread code and the renderer code. I was able to just rip out all the webpack configs and they didn't need to be replaced with anything (because it actually is zero-config). 

The way it builds the renderer now is that it targets `renderer/index.html` as the "entry-point" (instead of `index.js`) and _everything_ gets copied to the `dist/` folder now. The index.html page, the javascript, the CSS and the images.

I had to make some coding changes to make this work with Babel 7 as well.


.babelrc
```diff
@@ -1,22 +1,17 @@
...
+      "plugins": ["transform-class-properties"],
     },
     "test": {
-      "plugins": [
-        ["webpack-loaders", { "config": "webpack.config.node.js", "verbose": false }]
-      ]
+      "presets": ["@babel/env"],
     }
   }
 }
```
dpgraham: ParcelJS is smart about knowing which presets/plugins are needed and using those without having to configure `.babelrc`.

.gitignore
```diff
@@ -4,8 +4,10 @@ dist
 release
 main.js
 main.js.map
+main.map
 
 *.log
 .DS_Store
 
 Certificates.p12
+.cache/
```
dpgraham: This is the `.cache` folder that Parcel uses to cache modules for faster builds.

.vscode/launch.json
```diff
@@ -14,6 +14,20 @@
...
+        "request": "launch",
+        "program": "${workspaceFolder}/node_modules/mocha/bin/_mocha",
+        "args": [
+          "--timeout", "999999",
+          "--colors",
+          "${file}",
+          "--require", "@babel/register",
+        ],
+        "console": "integratedTerminal",
+        "internalConsoleOptions": "neverOpen"
```
dpgraham: This is unrelated to this PR. Just added VSCode build so we can debug mocha tests.

app/main/appium-method-handler.js
```diff
@@ -214,41 +214,40 @@ export default class AppiumMethodHandler {
...
-    log.info(`The following session window ID's have active sessions: '${JSON.stringify(_.keys(appiumHandlers))}'`);
-    return handler;
-  }
-
-  static killSession (winId, killedByUser) {
-    const {appiumHandlers} = AppiumMethodHandler;
-    const handler = appiumHandlers[winId];
-    delete AppiumMethodHandler.appiumHandlers[winId];
-    log.info(`Killing session for window with id: ${winId}`);
+export function createSession (driver, sender, winId) {
```
dpgraham: Had to change from `static` to `export` because Babel 7 doesn't accept the format of exporting objects and then doing destructured imports. You have to explicitly export each named variable.

app/main/auto-updater/config.js
```diff
@@ -1,12 +1,15 @@
...
-  baseFeedUrl: `https://hazel-server-pxufsrwofl.now.sh`,
-  getFeedUrl (version) {
-    let platform = process.platform;
-    if (platform.toLowerCase() === 'linux') {
-      platform = 'AppImage';
-    }
-    return `${config.baseFeedUrl}/update/${platform}/${version}`;
+const baseFeedUrl = `https://hazel-server-pxufsrwofl.now.sh`;
+
+export function getFeedUrl (version) {
```
dpgraham: Same as my comment from 'appium-method-handler'. Babel7 issue.

app/main/auto-updater/index.js
```diff
@@ -13,10 +13,11 @@ import _ from 'lodash';
 import env from '../../env';
 
 const isDev = process.env.NODE_ENV === 'development';
+const runningLocally = isDev || process.env.RUNNING_LOCALLY;
   
 let checkNewUpdates = _.noop;
 
-if (!isDev) {
+if (!isDev && !runningLocally) {
```
dpgraham: The script `npm start` runs the unpackaged Electron app, which is unsigned. If it's unsigned, trying to run `autoUpdate` will cause the app to break, because autoUpdate requires apps to be signed. 

Add this flag to skip auto update so we can still test unpackaged, unsigned builds.

app/main/helpers.js
```diff
@@ -25,7 +25,7 @@ export function openBrowserWindow (route, opts) {
   // note that __dirname changes based on whether we're in dev or prod;
   // in dev it's the actual dirname of the file, in prod it's the root
   // of the project (where main.js is built), so switch accordingly
-  let htmlPath = path.resolve(__dirname,  isDev ? '..' : 'app', 'renderer', 'index.html');
+  let htmlPath = path.resolve(__dirname,  'dist', 'index.html');
```
dpgraham: Since parcel uses `index.html` as it's entry point, `index.html` now gets copied over to `dist/`, so we use that file instead.

app/main/main.development.js
```diff
@@ -61,7 +58,7 @@ app.on('ready', async () => {
     minHeight: 600,
   });
 
-  mainWindow.loadURL(`file://${indexPath}/renderer/index.html`);
+  mainWindow.loadURL(`file://${__dirname}/dist/index.html`);
```
dpgraham: ditto my `helpers.js` comment

app/renderer/components/.postcssrc
```diff
@@ -0,0 +1,3 @@
+{
+  "modules": true
+}
```
dpgraham: This tells Parcel that any CSS file inside `app/renderer/components` is loaded using CSS modules... anything outside of this folder is just raw CSS.

app/renderer/components/ServerMonitor/ServerMonitor.js
```diff
@@ -5,6 +5,8 @@ import { STATUS_RUNNING, STATUS_STOPPING,
 import styles from './ServerMonitor.css';
 import AnsiConverter from 'ansi-to-html';
 
+import AppiumSmallMagenta from '../../images/appium_small_magenta.png';
```
dpgraham: Parcel handles importing of images by returning a path to that image. The file located at `app/renderer/images/appium_small_magenta.png` will be copied over to `dist/` and this module will load that file.

app/renderer/index.html
```diff
@@ -5,25 +5,11 @@
...
   <body>
     <div id="root"></div>
-    <script>
-      (function() {
-        const script = document.createElement('script');
-        script.src = (process.env.HOT) ? 'http://localhost:3000/dist/bundle.js' : '../../dist/bundle.js';
-        document.write(script.outerHTML);
-      }());
-    </script>
+    <script type="text/javascript" src="index.js"></script>
```
dpgraham: `/renderer/index.html` is now the entry point and Parcel looks for `index.js` as the JS entrypoint for this script tag and builds at as `<script type="text/javascript" src="renderer.<cache-bust>.js"></script>`

app/renderer/index.html
```diff
@@ -5,25 +5,11 @@
...
-        if (!process.env.HOT) {
-          const link = document.createElement('link');
-          link.rel = 'stylesheet';
-          link.href = '../../dist/style.css';
-          document.write(link.outerHTML);
-        }
-      }());
-    </script>
+    <link rel="stylesheet" href="styles/app.global.css" />
+    <link rel="stylesheet" href="styles/github-gist-theme.global.css" />
```
dpgraham: These two files get loaded as raw css.

app/renderer/index.js
```diff
@@ -6,35 +6,44 @@ import { syncHistoryWithStore } from 'react-router-redux';
...
+    </Provider>,
+    document.getElementById('root')
+  );
+}
+
+renderApp();
+
+if (module.hot) {
+  module.hot.accept(renderApp);
+}
```
dpgraham: This is the Parcel hotloading of React.

app/renderer/util.js
```diff
@@ -42,7 +43,7 @@ export function getOptimalXPath (doc, domNode, uniqueAttributes = ['id']) {
     // If this node has siblings of the same tagName, get the index of this node
     if (domNode.parentNode) {
       // Get the siblings
-      const childNodes = [...domNode.parentNode.childNodes].filter((childNode) => (
+      const childNodes = Array.prototype.slice.call(domNode.parentNode.childNodes, 0).filter((childNode) => (
```
dpgraham: Babel 7 incompatibility.... doesn't allow the spread operator on non-arrays.

package.json
```diff
@@ -21,27 +21,37 @@
   "homepage": "https://github.com/appium/appium-desktop",
   "main": "main.js",
   "scripts": {
+    "//": "TEST SCRIPTS",
     "test": "npm run lint && npm run test-unit && npm run test-integration",
-    "test-e2e": "cross-env FORCE_NO_WRONG_FOLDER=true RUNNING_IN_SPECTRON=true mocha --timeout 600000 --reporter mocha-multi-reporters --reporter-options configFile=./test/mochareporters.json --require ./node_modules/babel-polyfill/dist/polyfill.min.js --require babel-core/register ./test/e2e",
```
dpgraham: Moved `--require` option to `mocha.opts`.

package.json
```diff
@@ -21,27 +21,37 @@
...
+
+    "//": "PRODUCTION BUILD SCRIPTS",
+    "build-main": "cross-env NODE_ENV=production parcel build --target node --out-dir . --out-file main.js ./app/main/main.development.js",
+    "build-renderer": "cross-env NODE_ENV=production parcel build --out-dir dist --target electron --public-url . ./app/renderer/index.html",
+    "build": "npm run clean && npm run build-main && npm run build-renderer",
+    "start": "cross-env RUNNING_LOCALLY=true FORCE_NO_WRONG_FOLDER=true electron ./",
+
+    "//": "DEV BUILD SCRIPTS",
+    "watch-main": "cross-env NODE_ENV=developpment parcel --target node --out-dir . --out-file main.js ./app/main/main.development.js",
+    "watch-renderer": "cross-env NODE_ENV=developpment parcel --out-dir dist --target electron --public-url . ./app/renderer/index.html",
```
dpgraham: ^ These two scripts build the 'main' and 'renderer' thread and run a server that rebuilds on the fly.

`public-url` makes it relative instead of absolute so that `dist/` isn't treated as the root.

package.json
```diff
@@ -21,27 +21,37 @@
...
-    "clean": "rimraf node_modules && npm install",
-    "start": "cross-env NODE_ENV=production electron ./",
-    "start-hot": "cross-env HOT=1 NODE_ENV=development electron -r ./node_modules/babel-register -r ./node_modules/babel-polyfill --inspect=5858 ./app/main/main.development --color",
-    "postinstall": "node node_modules/fbjs-scripts/node/check-dev-engines.js package.json && rimraf node_modules/jimp/browser",
-    "dev": "concurrently --kill-others \"npm run hot-server\" \"npm run start-hot\" --color",
+
+    "//": "PRODUCTION BUILD SCRIPTS",
+    "build-main": "cross-env NODE_ENV=production parcel build --target node --out-dir . --out-file main.js ./app/main/main.development.js",
+    "build-renderer": "cross-env NODE_ENV=production parcel build --out-dir dist --target electron --public-url . ./app/renderer/index.html",
+    "build": "npm run clean && npm run build-main && npm run build-renderer",
```
dpgraham: Cleans the `dist/` folder, builds the main NodeJS build and the renderer.

package.json
```diff
@@ -21,27 +21,37 @@
...
+    "start": "cross-env RUNNING_LOCALLY=true FORCE_NO_WRONG_FOLDER=true electron ./",
+
+    "//": "DEV BUILD SCRIPTS",
+    "watch-main": "cross-env NODE_ENV=developpment parcel --target node --out-dir . --out-file main.js ./app/main/main.development.js",
+    "watch-renderer": "cross-env NODE_ENV=developpment parcel --out-dir dist --target electron --public-url . ./app/renderer/index.html",
+    "start-dev": "cross-env RUNNING_LOCALLY=true FORCE_NO_WRONG_FOLDER=true NODE_ENV=development electron ./",
+    "dev": "cross-env NODE_ENV=development concurrently --kill-others \"npm run watch-main\" \"npm run watch-renderer\" \"npm run start-dev\" ",
```
dpgraham: Builds the main and renderer concurrently and runs Electron.

test/e2e/main-e2e.test.js
```diff
@@ -27,6 +27,9 @@ before(async function () {
   this.timeout(process.env.TRAVIS || process.env.APPVEYOR ? 10 * 60 * 1000 : 30 * 1000);
   this.app = new Application({
     path: appPath,
+    env: {
+      FORCE_NO_WRONG_FOLDER: true,
```
dpgraham: Don't give the 'Not in Application Folder' error for E2E tests....

Don't know how this worked before.

test/e2e/pages/inspector-page-object.js
```diff
@@ -65,11 +65,11 @@ InspectorPage.selectors = {
   customServerPort: '#customServerPort',
   addDesiredCapabilityButton: '#btnAddDesiredCapability',
   formSubmitButton: '#btnStartSession',
-  inspectorToolbar: 'div[class*=Inspector__inspector-toolbar]',
+  inspectorToolbar: 'div[class*=_inspector-toolbar]',
```
dpgraham: Parcel does style imports differently, but still works the same.

package.json
```diff
@@ -69,8 +71,6 @@
     "files": [
       "dist/",
       "node_modules/**/*",
-      "app/renderer/index.html",
-      "app/renderer/images/",
```
dpgraham: Images in index.html get built to `dist/` now... everything in `/app` can be ignored.

app/main/appium-method-handler.js
jlipps: 👍

app/main/helpers.js
jlipps: finally, i hated this hack

jlipps: looks like we can do away with the comment above, now?

jlipps: wait so `__dirname` is always the root of the project now, not the directory of the current file?

app/renderer/components/ServerMonitor/ServerMonitor.js
jlipps: so parceljs has an image import loader or something that dynamically updates the `AppiumSmallMagenta` name based on the build?

app/main/helpers.js
dpgraham: `helpers.js` gets bundled into `main.js`, which runs in the root.

dpgraham: It used to be that either the main thread would be run from `main/main.development` for `dev` and `main.js` from `prod`. Now I just have always build `main.js` (because it's fast, and the performance isn't an issue).

app/renderer/components/ServerMonitor/ServerMonitor.js
dpgraham: It copies `images/appium-small-magenta.png` to `dist/` (along with the other assets )and renames it `appium-small-magenta.<cache-bust>.png`.

When `ServerMonitor.js` imports `AppiumSmallMagenta` it imports the name of that file as a string that can be passed into the the `src` attribute of ``

Issue #818, opened 2019-02-08
Localise the application interface


mykola-mokhnach:
There is [i18next](https://www.i18next.com) library, which might be handy for localization purposes. We could start with adding one locale first. Crowdin.com could be used as TMS (free for open source).

[Here](https://phraseapp.com/blog/posts/building-an-electron-app-with-internationalization-i18n/) is the tutorial on how to perform the initial setup.

KazuCocoa:
Will work for 🇯🇵

dpgraham:
Closed as completed.

PR #819, opened 2019-02-09
Perform the initial setup for the localization

mykola-mokhnach:
Addresses https://github.com/appium/appium-desktop/issues/818

de


app/configs/i18next.config.client.js
```diff
@@ -0,0 +1,28 @@
+import i18next from 'i18next';
```
dpgraham: I'd rather this file be called `i18next.config.renderer.js`

locales/en/translation.json
```diff
@@ -0,0 +1,38 @@
+{
+  "en": "en",
```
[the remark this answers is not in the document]
mykola-mokhnach: yes, but later

mykola-mokhnach:
main_win

dpgraham:
Awesome!

app/configs/i18next.config.client.js
mykola-mokhnach: fixed

PR #844, opened 2019-03-04
feat: Run commands from Appium Desktop GUI

dpgraham:
* Add a tab that lets user navigate to `Actions` tab (default is source)
* In this tab, user can select a command and then execute that command
* Command definitions are defined in `Inspector/shared.js`
* If the command has args, it pops a modal that lets user input fields
* If the command has no args, it runs the command when pressing button
* When a command is executed, if the command has a non-null result, display that result as a notification

Other:
* Fixed problems with the error notifications not producing very helpful error messages
* Renamed some of the CSS to be less specific

Screencaps:
screen shot 2019-03-04 at 1 53 46 pm

screen shot 2019-03-04 at 1 56 20 pm

screen shot 2019-03-04 at 1 57 00 pm


jlipps:
Love this idea! Before looking at code, one high level thought:

Why put the actions palette as a tabbed interface alongside source? Why not a toolbar button that opens up a modal? I don't love the way that the top of that new tab bar pushes the content down so it no longer aligns with the left and right columns.

I'd even be happy if somehow the new tab interface was part of the title bar of the white card, so that at least the lines went across well.

dpgraham:
@jlipps what do you think of it looking like this?

screen shot 2019-03-04 at 4 35 52 pm

app/renderer/actions/Session.js
```diff
@@ -99,7 +99,7 @@ export function showError (e, methodName, secs = 5) {
   }
 
   notification.error({
-    message: 'Error',
+    message: methodName ? `Call to '${methodName}' failed` : 'Error',
```
mykola-mokhnach: use i18next to translate these strings and move them into resources

app/renderer/actions/shared.js
```diff
@@ -8,8 +11,23 @@ export function bindClient () {
...
+    let {res, methodName} = resp;
+
+    // Ignore empty objects
+    if (_.isObject(res) && _.isEmpty(res)) {
+      res = null;
+    }
+
+    if (!_.isNull(res) && !_.isUndefined(res)) {
+      notification.success({
+        message: `'${methodName}' ${i18n.t('result')}`,
```
mykola-mokhnach: It's better to use a template string there. In some languages the order of words might be different

app/renderer/actions/shared.js
```diff
@@ -8,8 +11,23 @@ export function bindClient () {
...
+        description: `${i18n.t('Command was returned with result')}: '${_.truncate(JSON.stringify(res), {length: 2000})}'`,
```
mykola-mokhnach: same here - the whole string must be represented as a template

app/renderer/actions/shared.js
```diff
@@ -8,8 +11,23 @@ export function bindClient () {
...
   ipcRenderer.on('appium-client-command-response', (evt, resp) => {
     // Rename 'id' to 'elementId'
+    let {res, methodName} = resp;
+
+    // Ignore empty objects
+    if (_.isObject(res) && _.isEmpty(res)) {
+      res = null;
+    }
+
+    if (!_.isNull(res) && !_.isUndefined(res)) {
```
mykola-mokhnach: util.hasValue

app/renderer/actions/shared.js
```diff
@@ -22,10 +40,11 @@ export function bindClient () {
    * If we hear back with an error, reject the promise
    */
   ipcRenderer.on('appium-client-command-response-error', (evt, resp) => {
-    const {e, uuid} = resp;
+    let {e, uuid} = resp;
+    e = JSON.parse(e);
```
mykola-mokhnach: does it work if an error object being serialized like this?

app/renderer/components/Inspector/Actions.js
```diff
@@ -0,0 +1,98 @@
...
+
+  executeCommand () {
+    const { pendingAction, cancelPendingAction, applyClientMethod, t } = this.props;
+    let {args, action} = pendingAction;
+    let {methodName} = action;
+
+    // Special case for 'startActivity'
+    // TODO: Fix these... args aren't getting through
+    if (action.methodName === 'startActivity') {
+      args = {appPackage: args[0], appActivity: args[1], appWaitPackage: args[2],
```
mykola-mokhnach: cannot we pass args as a dictionary?

app/renderer/components/Inspector/Actions.js
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    // Special case for 'execute'
+    if (action.methodName === 'execute') {
+      if (!_.isEmpty(args[1])) {
+        try {
+          args[1] = JSON.parse(args[1]);
+        } catch (e) {
+          notification.error({
+            message: `${t('Invalid JSON')}: args[1]}`,
```
mykola-mokhnach: use template string

app/renderer/components/Inspector/Actions.js
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  render () {
+    const {
+      t, selectActionGroup, selectSubActionGroup, selectedActionGroup, selectedSubActionGroup, pendingAction,
+      cancelPendingAction, setActionArg } = this.props;
+    return [
+      <Row gutter={16} style={{marginBottom: '8px'}}>
+        <Col span={24}>
+          <Select style={{width: '100%'}} onChange={(actionGroupName) => selectActionGroup(actionGroupName)} placeholder="Select Action Group">
```
mykola-mokhnach: why width is not in css?

mykola-mokhnach: placehodler text should be translated

dpgraham: I'll make a note to move all the style tags to css rules for tomorrow.

app/renderer/components/Inspector/Actions.js
```diff
@@ -0,0 +1,98 @@
...
+            { _.keys(actionDefinitions).map((actionGroup) => <Option key={actionGroup}>{t(actionGroup)}</Option>) }
+          </Select>
+        </Col>
+      </Row>,
+      selectedActionGroup && <Row>
+        <Col span={24}>
+          <Select style={{width: '100%'}} onChange={(actionGroupName) => selectSubActionGroup(actionGroupName)} placeholder="Select Sub Group">
```
mykola-mokhnach: placeholder text should be translated

app/renderer/components/Inspector/Inspector.js
```diff
@@ -63,18 +68,29 @@ export default class Inspector extends Component {
...
         {showRecord &&
           <RecordedActions {...this.props} />
         }
-        <Card
-          title={<span><Icon type="file-text" /> App Source</span>}
-          className={InspectorStyles['source-tree-card']}>
-          <Source {...this.props} />
-        </Card>
+        <Tabs activeKey={selectedInteractionMode} size="small" onChange={(tab) => selectInteractionMode(tab)}>
+          <TabPane tab="Source" key={INTERACTION_MODE.SOURCE}>
```
mykola-mokhnach: are tab names translatable?

app/renderer/components/Inspector/RecordedActions.js
```diff
@@ -89,7 +89,7 @@ class RecordedActions extends Component {
 
     const highlightedCode = this.code(false);
 
-    return <Card title={<span><Icon type="code-o"/>{t('Recorder')}</span>}
+    return <Card title={<span><Icon type="code-o"/> {t('Recorder')}</span>}
```
mykola-mokhnach: is that space needed?

dpgraham: It is, yeah. It looks weird if there's no space between it and the icon.

app/renderer/components/Inspector/shared.js
```diff
@@ -52,3 +52,115 @@ export const SCREENSHOT_INTERACTION_MODE = {
...
+
+export const actionArgTypes = {
+  STRING: 'string',
+  NUMBER: 'number',
+};
+
+const { STRING, NUMBER } = actionArgTypes;
+
+export const actionDefinitions = {
+  'Device': {
```
mykola-mokhnach: should these action definition captions be translated?

app/renderer/components/Stub.js
```diff
@@ -3,10 +3,10 @@ import React, { Component } from 'react';
 export default class Stub extends Component {
 
   render () {
-    const { someProp } = this.props;
+    const { someProp, t } = this.props;
 
     return <div>
-      Hello {someProp}!
+      {t('Hello')} {someProp}!
```
mykola-mokhnach: the whole string should be a template

app/renderer/components/Inspector/shared.js
dpgraham: Not here, but inside the React component, yes.

app/renderer/components/Stub.js
dpgraham: This is just a stub for reference purpose. But I can change it anyways.

app/renderer/components/Inspector/Inspector.js
dpgraham: Yeah I can do `<TabPane tab=${t('source')} />`

app/renderer/lib/client-frameworks/robot.js
```diff
@@ -97,6 +97,214 @@ ${this.indent(code, 4)}
   codeFor_swipe (varNameIgnore, varIndexIgnore, x1, y1, x2, y2) {
     return `Swipe    ${x1}    ${y1}    ${x2}    ${y2}`;
   }
+
+  codeFor_getCurrentActivity () {
```
mykola-mokhnach: FXIME?

app/renderer/reducers/Inspector.js
```diff
@@ -333,6 +336,49 @@ export default function inspector (state = INITIAL_STATE, action) {
...
+          args: [],
+        }
+      };
+
+    case SET_ACTION_ARG:
+      return {
+        ...state,
+        pendingAction: {
+          ...state.pendingAction,
+          args: Object.assign([], state.pendingAction.args, {[action.index]: action.value}), // Replace 'value' at 'index'
```
mykola-mokhnach: What is done here?

mykola-mokhnach:
All the strings marked with `t` should must be added into `assets/locales/en/translation.json`. While adding them please make sure the JSON keys don't contain neither dots not colons. Use custom  identifiers for such strings as well as for template strings

app/renderer/reducers/Inspector.js
dpgraham: Makes a copy of `state.pendingAction.args` but replaces the value at `action.index` with `action.value`....

this was the best way I could think of to change a value of an array item immutably.

dpgraham:
@mykola-mokhnach I just added an eslint rule called `jsx-no-hardcoded-content` which validates that there's no hardcoded content in JSX. This will force us to use `t()` all the time.

app/renderer/reducers/Inspector.js
mykola-mokhnach: let say we have `const a = [1,2,3,]` then in order to replace the second element with 4 we could `const b = [...a.slice(0, 2), 4, ...s.slice(3)]`

app/renderer/components/Inspector/Actions.js
dpgraham: For the commands, `args` is an array. I didn't want users to have to input a JSON string so instead I let them put in an array of args that gets converted to JSON and then called on `wd`

jlipps:
@dpgraham that's better, though in this case i'd expected the 'selected element' card to disappear when we click on actions.

dpgraham:
@jlipps it does... `Selected Element` is part of the source tab now.

app/renderer/lib/client-frameworks/robot.js
dpgraham: I'll mention it to the person who put in the PR for the robot framework recording code.

app/renderer/actions/shared.js
dpgraham: The response is always a JSON object that's passed over the ipcRenderer

app/renderer/components/Inspector/LocatedElements.js
```diff
@@ -34,9 +34,9 @@ class LocatedElements extends Component {
 
     return <Row>
       <p className={InspectorStyles['back-link-container']}>
-        <a onClick={(e) => e.preventDefault() || clearSearchResults()}>&lt;&lt; Back</a>
+        <a onClick={(e) => e.preventDefault() || clearSearchResults()}>{t('back')}</a>
       </p>
-      Elements (<span>{locatedElements.length}</span>):
+      {t('elementsCount', {elementCount: locatedElements.length})}
```
mykola-mokhnach: You could still keep `span` tags if you want by using I18Trans component

assets/locales/en/translation.json
```diff
@@ -154,5 +154,147 @@
...
+  "couldNotObtainSource": "Could not obtain source: {{errorMsg}}",
+  "iOS": "iOS",
+  "Android": "Android",
+  "desiredCapabilitiesDocumentation": "Desired Capabilities Documentation",
+  "startSession": "Start Session",
+  "saveAs": "Save As...",
+  "attachToSession": "Attach to Session",
+  "Hello": "Hello",
+  "selectedElement": "Selected Element",
+  "back": "&lt;&lt; Back",
```
mykola-mokhnach: I assume these must be escaped by i18next-react automatically, so you could put `<<` instead

assets/locales/en/translation.json
```diff
@@ -154,5 +154,147 @@
...
+  "filepath": "filepath",
+  "connectToExistingSessionInstructions": "If you have an already-running session of the above server type, you can attach an inspector to it directly.",
+  "selectSessionIDInDropdown": "Select the Session ID in the dropdown below.",
+  "allowUnauthorizedCerts": "Allow Unauthorized Certificates",
+  "Use Proxy": "Use Proxy",
+  "Automatic Server": "Automatic Server",
+  "Custom Server": "Custom Server",
+  "Desired Capabilities": "Desired Capabilities",
+  "Saved Capability Sets": "Saved Capability Sets {{savedSessionsCount}}",
+  "Attach to Session": "Attach to Session...",
```
mykola-mokhnach: Please use Unicode ellipsis instead of three dots

mykola-mokhnach: similar comment to other occurrences

assets/locales/en/translation.json
```diff
@@ -154,5 +154,147 @@
...
+  "localhost": "localhost",
+  "Host": "Host",
+  "Port": "Port",
+  "Preset name": "Preset name",
+  "usingDataFoundIn": "Using data found in ${{environmentVariable}}",
```
mykola-mokhnach: do we need the dollar sign?

app/renderer/components/Session/AttachToSession.js
```diff
@@ -14,13 +14,17 @@ function formatCaps (caps) {
 
 export default class AttachToSession extends Component {
 
+  getSessionInfo (session) {
+    return `${session.id} -- ${formatCaps(session.capabilities)}`;
```
mykola-mokhnach: Please use Unicode dash

.eslintrc
```diff
@@ -15,6 +15,8 @@
     "import/no-unresolved": [2, {"ignore": ["electron"]}],
     "babel/arrow-parens": 0,
     "require-await": 2,
-    "shopify/jsx-no-hardcoded-content": 2
+    "shopify/jsx-no-hardcoded-content": [2, {
+      "checkProps": ["tab", "placeholder", "addonBefore", "addonAfter", "title"]
```
mykola-mokhnach: 👍

dpgraham:
Fixes pushed

dpgraham:
@jlipps would you have time to help out with the CSS? The scrolling of the Selected Element is broken and I'm pretty stumped trying to figure out what the problem is.

Also, I was able to change the CSS in the source so it uses native scroll instead of the react scroll component that I developed.

app/renderer/components/Inspector/Inspector.js
```diff
@@ -63,24 +67,37 @@ export default class Inspector extends Component {
...
-        <Card
-          title={<span><Icon type="tag-o" /> Selected Element</span>}
-          className={InspectorStyles['selected-element-card']}>
-          {path && <SelectedElement {...this.props}/>}
-          {!path && <i>{t('selectElementInSource')}</i>}
-        </Card>
+        <Tabs activeKey={selectedInteractionMode} size="small" onChange={(tab) => selectInteractionMode(tab)}>
+          <TabPane tab={t('Source')} key={INTERACTION_MODE.SOURCE}>
+            <div style={{display: 'flex'}}>
+              <div style={{flex: '1 1 0', width: 0}}>
```
mykola-mokhnach: hardcoded styles?

PR #863, opened 2019-03-13
chore: Fix ENV secret problem

dpgraham:
This pipeline runs `npm run crowdin-update` on every commit to master

### DIFF
--- a/ci-jobs/crowdin.yml
+++ b/ci-jobs/crowdin.yml
@@ -4,6 +4,9 @@
 # https://docs.microsoft.com/vsts/pipelines/languages/javascript
 jobs:
 - job: crowdin
+  variables:
+    CROWDIN_PROJECT_ID: $(CROWDIN_PROJECT_ID)
+    CROWDIN_PROJECT_KEY: $(CROWDIN_PROJECT_KEY)
   steps:
     - task: NodeTool@0
       inputs:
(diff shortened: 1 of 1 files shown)


ci-jobs/crowdin.yml
```diff
@@ -10,4 +10,6 @@ jobs:
         versionSpec: '11.x'
     - script: npm install
       displayName: NPM install
-    - script: npm run crowdin-update
\ No newline at end of file
+    - script: npm run crowdin-update
```
mykola-mokhnach: it automatically picks up these vars from the Travis settings and it worked before. Why do we need to change it?

dpgraham: This is azure pipelines. Secret variables need to be placed inside `env:`

mykola-mokhnach: this script actually requires both project name and key variables to be defined

mykola-mokhnach: also, it is only expected to be executed on merge to master and not for each commit

dpgraham: I have it set to only trigger on master commits and the project name gets passed into env because it isnt a secret variable. AP has a rule about requiring explicitly passing secret variables to individual tasks.

dpgraham:
Here's a passing test, for reference: https://dev.azure.com/appiumci/Appium%20CI/_build/results?buildId=1095

ci-jobs/crowdin.yml
dpgraham: *is a secret variable

PR #874, opened 2019-03-18
fix: inspector CSS

dpgraham:
The last commit to add the `Actions` tab messed up the CSS. This fixes it all.

Screen Shot 2019-03-18 at 1 16 28 PM

### DIFF
--- a/app/renderer/components/Inspector/Inspector.css
+++ b/app/renderer/components/Inspector/Inspector.css
@@ -66,6 +66,14 @@
     flex-flow: column;
 }
 
+.inspector-main .interaction-tab-container :global(.ant-tabs) {
+    height: 100%;
+}
+
+.inspector-main .interaction-tab-container :global(.ant-tabs-content) {
+    height: 100%;
+}
+
 .inspector-main .interaction-tab-container :global(.action-row) {
     display: flex;
 }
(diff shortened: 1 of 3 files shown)


jlipps:
i think our strategy so far has been not to show scroll bars, whereas i see scroll bars midway down the source box?

dpgraham:
@jlipps yeah I wanted it in the "Source" for horizontal scrolling.

jlipps:
@dpgraham does the horizontal scroll bar show up even if not in the middle of a scroll? if so, then it's against our design patterns so far.

also, it seems like the height of the inner content is not 100%, i'd expect to see the scroll bar at the bottom of the box, even if the content doesn't fill the whole box.

dpgraham:
Screen Shot 2019-03-21 at 11 18 44 AM

I've fixed it @jlipps .... the problem was I was setting `overflow: auto` to all `ul` trees. Just committed a change that makes it only do overflow for the root.

It only shows the horizontal and vertical scrollbars if it's overflowing. Otherwise it's hidden. If there was a way to make it vertically scrollable without showing the scrollbar but horizontally scrollable with showing the scrollbar, that would be ideal, but not sure if that's possible.

We definitely should make the horizontal scrollbar visible though because I wouldn't count people having a mouse that scrolls horizontally.

dpgraham:
@jlipps what are your thoughts on how this screenshot looks ^

dpgraham:
I'm going to merge this because I need to get a release out soon. But if you want to change it, I'm open to putting in another PR.

PR #876, opened 2019-03-19
Don't show all cloud providers in top menu.

dpgraham:
* Too many cloud providers, getting out of hand
* New way is to only show `Automatic | Custom | Add Cloud Provider +`
Screen Shot 2019-03-19 at 4 23 53 PM
* Selecting `Add Cloud Provider +` brings up a modal with a grid of cloud providers to choose from
Screen Shot 2019-03-19 at 4 24 04 PM

* Removed the old method of selecting a cloud provider from a menu to show which ones you want to be whitelisted as "visible"
* Made a file `CloudProviders.js` that holds all the info about cloud providers


jlipps:
One usability question. Would it be better if instead it was "Select Cloud Providers" and each one was togglable? That way if you want to hide some again it's easy?

dpgraham:
@jlipps that's a good idea actually. Probably better than having the `x` next to it.

app/renderer/actions/Session.js
```diff
@@ -635,4 +634,41 @@ function removeNewSessionListeners () {
...
+export function stopAddCloudProvider () {
+  return (dispatch) => {
+    dispatch({type: IS_ADDING_CLOUD_PROVIDER, isAddingProvider: false});
+  };
+}
+
+export function addVisibleProvider (provider) {
+  return async (dispatch, getState) => {
+    let currentProviders = getState().session.visibleProviders;
+    const providers = _.union(currentProviders, [provider]);
```
mykola-mokhnach: [... currentProviders, provider]

assets/locales/en/translation.json
```diff
@@ -310,5 +310,7 @@
   "commandWasReturnedWithResult": "Command was returned with result: {{result}}",
   "Invalid JSON": "Invalid JSON: {{json}}",
   "TestingBot Key": "TestingBot Key",
-  "TestingBot Secret": "TestingBot Secret"
+  "TestingBot Secret": "TestingBot Secret",
+  "Select Cloud Provider": "Select Cloud Provider",
+  "Add Cloud Provider": "Add Cloud Provider +"
```
mykola-mokhnach: Do we need this plus?

app/renderer/actions/Session.js
```diff
@@ -48,19 +50,16 @@ export const SAVE_RAW_DESIRED_CAPS = 'SAVE_RAW_DESIRED_CAPS';
...
-  kobiton: 'kobiton',
-  perfecto: 'perfecto',
-  pcloudy: 'pcloudy',
-  testingbot: 'testingbot'
-};
+export const IS_ADDING_CLOUD_PROVIDER = 'IS_ADDING_CLOUD_PROVIDER';
+
+export const SET_PROVIDERS = 'SET_PROVIDERS';
+
+const serverTypes = {};
```
mykola-mokhnach: = CloudProviders.reduce ...

app/renderer/actions/Session.js
```diff
@@ -3,9 +3,10 @@ import settings from '../../shared/settings';
 import { v4 as UUID } from 'uuid';
 import { push } from 'connected-react-router';
 import { notification } from 'antd';
-import { debounce, toPairs } from 'lodash';
+import _, { debounce, toPairs } from 'lodash';
```
mykola-mokhnach: I'd rather use `_` syntax or named imports. using both is confusing

app/renderer/components/Session/CloudProviderSelector.js
```diff
@@ -0,0 +1,63 @@
...
+import _ from 'lodash';
+import { Modal, Row, Col, Button } from 'antd';
+import CloudProviders from './CloudProviders';
+import SessionStyles from './Session.css';
+
+export default class CloudProviderSelector extends Component {
+
+  constructor (props) {
+    super(props);
+    this.state = {
```
mykola-mokhnach: I think there is a shortcut to this, so you could simply declare state without declaring the constructor.

app/renderer/components/Session/CloudProviderSelector.js
```diff
@@ -0,0 +1,63 @@
+import React, { Component } from 'react';
+import _ from 'lodash';
+import { Modal, Row, Col, Button } from 'antd';
+import CloudProviders from './CloudProviders';
+import SessionStyles from './Session.css';
+
+export default class CloudProviderSelector extends Component {
```
mykola-mokhnach: Is i18n added automatically to props? I thought you need to wrap the exported class into i18n HOC before exporting it to make that possible

app/renderer/components/Session/Session.js
```diff
@@ -44,124 +24,113 @@ export default class Session extends Component {
...
 
-  async checkProvidersVisibility () {
-    const visibleProviders = {};
-    for (let [providerName, provider] of _.toPairs(CloudProviders)) {
-      if (await provider.isVisible()) {
-        visibleProviders[providerName] = true;
-      }
+  handleSelectServerTab (tab) {
+    const {changeServerType, addCloudProvider} = this.props;
+    if (tab === 'addCloudProvider') {
```
mykola-mokhnach: can it be a constant?

app/renderer/components/Session/Session.js
```diff
@@ -44,124 +24,113 @@ export default class Session extends Component {
...
+      addCloudProvider();
+      return;
     }
-    this.setState({
-      ...this.state,
-      visibleProviders
-    });
+    changeServerType(tab);
```
mykola-mokhnach: why do we call the entity ServerType when changing and CloudProvided when adding?

app/renderer/components/Session/CloudProviderSelector.js
dpgraham: The props get passed from the root to all subcomponents.

app/renderer/components/Session/Session.js
dpgraham: `serverType` encompasses Cloud Providers and local

mykola-mokhnach:
it seems like CI is not quite happy

dpgraham:
Screen Shot 2019-03-21 at 10 36 00 AM

@jlipps ^

app/renderer/components/Session/CloudProviderSelector.js
dpgraham: This is how `i18n` gets passed into the component:

```js
<CloudProviderSelector {...this.props} />
```

app/renderer/actions/Session.js
```diff
@@ -48,19 +50,18 @@ export const SAVE_RAW_DESIRED_CAPS = 'SAVE_RAW_DESIRED_CAPS';
...
-  kobiton: 'kobiton',
-  perfecto: 'perfecto',
-  pcloudy: 'pcloudy',
-  testingbot: 'testingbot'
-};
+export const IS_ADDING_CLOUD_PROVIDER = 'IS_ADDING_CLOUD_PROVIDER';
+
+export const SET_PROVIDERS = 'SET_PROVIDERS';
+
+const serverTypes = {};
```
mykola-mokhnach:
```
const serverTypes = _.keyBy([...keys(CloudProviders), 'local', 'remote'], (x) => x))
```

PR #890, opened 2019-03-21
Use triager bot to update crowdin

dpgraham:
To update crowdin translations, we need to run `npm run crowdin-sync`. Instead of remembering to do that, I'd rather have a bot do this once a week and make a [pull request](https://github.com/appium/appium-desktop/pull/889)


dpgraham:
I'm glad you like it. We do something similar with Appium.io document generation. Less things to remember is always nice.

PR #911, opened 2019-03-28
fix: Parse JSON for updateSettings

dpgraham:
Need to parse 'settings' JSON before passing it through to wd

(cc: @wswebcreation)

### DIFF
--- a/app/renderer/components/Inspector/Actions.js
+++ b/app/renderer/components/Inspector/Actions.js
@@ -51,6 +51,20 @@ export default class Actions extends Component {
       }
     }
 
+    // Special case for 'updateSettings'
+    if (action.methodName === 'updateSettings') {
+      if (_.isString(args[0])) {
+        try {
+          args[0] = JSON.parse(args[0]);
+        } catch (e) {
+          notification.error({
+            message: t('invalidJson', {json: args[1]}),
+            duration: 5,
+          });
+        }
+      }
+    }
+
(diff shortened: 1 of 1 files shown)


app/renderer/components/Inspector/Actions.js
```diff
@@ -51,6 +51,20 @@ export default class Actions extends Component {
       }
     }
 
+    // Special case for 'updateSettings'
+    if (action.methodName === 'updateSettings') {
+      if (_.isString(args[0])) {
```
mykola-mokhnach: this if could be combined with the previous one using &&

app/renderer/components/Inspector/Actions.js
```diff
@@ -51,6 +51,20 @@ export default class Actions extends Component {
...
+        try {
+          args[0] = JSON.parse(args[0]);
+        } catch (e) {
+          notification.error({
+            message: t('invalidJson', {json: args[1]}),
```
mykola-mokhnach: do we know that we have at least 2 args?

mykola-mokhnach: or it should be args[0]?

dpgraham: Good catch, thanks.

wswebcreation:
Hi @dpgraham 

Thanks for this fix. Is this already in the latest release? I tried this with the 1.12.0 and it still fails

PR #813, opened 2019-01-23
Add way to select overlapped behind views of Screenshot component

muyoungko:
Problems I have experienced is like following screen shot.

So I made way to select on screenshot component.

PR #917, opened 2019-03-29
Revert "Add way to select overlapped behind views of Screenshot component"

dpgraham:
Reverts appium/appium-desktop#813

@muyoungko I'm going to have revert this, sadly. I was testing it today and I had some cases where I was mousing over menu items and it was selecting the entire app Window.


dpgraham:
@muyoungko I would be happy to review another PR if you can fix it.

muyoungko:
Ok I got it, thanks.

PR #922, opened 2019-04-01
fix: Stop ignoring result for element location/size

dpgraham:
The 'ignoreResult' flag was causing the result to not be sent back to the renderer process. This was done because there's some cases where we don't want the result to be displayed using "notification" but this is a problem for highlighting an element because it wasn't getting the result.
This changes it so that 'ignoreResult' still sends the result back to the renderer but skips the notification if it's 'ignoreResult=true'.

### DIFF
--- a/app/renderer/actions/shared.js
+++ b/app/renderer/actions/shared.js
@@ -1,5 +1,6 @@
 import { ipcRenderer } from 'electron';
 import { notification } from 'antd';
+import { util } from 'appium-support';
 import i18n from '../../configs/i18next.config.renderer';
 import _ from 'lodash';
 import UUID from 'uuid';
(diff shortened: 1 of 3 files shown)


app/renderer/actions/shared.js
```diff
@@ -23,7 +23,7 @@ export function bindClient () {
 
     const truncatedResult = _.truncate(JSON.stringify(res), {length: 2000});
 
-    if (!_.isNull(res) && !_.isUndefined(res)) {
+    if (!_.isNull(res) && !_.isUndefined(res) && !resp.ignoreResult) {
```
mykola-mokhnach: util.hasValue(res) && !resp.ignoreResult

PR #979, opened 2019-05-09
make configurable get running sessions path

KazuCocoa:
Currently, `getRunningSessions` sends requests to `/wd/hub`.
Custom server tab and other services also can configure their own path instead of `/wd/hub`.
But then, this feature never listens to the paths.

image

In this PR, I changed to pass `path` and `ssl` preference to the request.
(And I understood `session.server.${serverType}.hostname` etc are used then..)

Can you review this? I wondered if this implementation had intentions.
@jlipps @dpgraham

### DIFF
--- a/app/main/appium.js
+++ b/app/main/appium.js
@@ -321,9 +321,9 @@ function connectClientMethodListener () {
 }
 
 const getCurrentSessions = _.debounce(async (evt, data) => {
-  const {host, port, path, ssl} = data;
+  const {host, port, path: appiumPath = '/wd/hub', ssl} = data;
   try {
-    const res = await request(`http${ssl ? 's' : ''}://${host}:${port}${path ? path : '/wd/hub'}/sessions`);
+    const res = await request(`http${ssl ? 's' : ''}://${host}:${port}${appiumPath}/sessions`);
     evt.sender.send('appium-client-get-sessions-response', {res});
   } catch (e) {
     evt.sender.send('appium-client-get-sessions-fail');
--- a/app/main/appium.js
+++ b/app/main/appium.js
@@ -321,9 +321,9 @@ function connectClientMethodListener () {
 }
 
 const getCurrentSessions = _.debounce(async (evt, data) => {
-  const {host, port, ssl} = data;
+  const {host, port, path, ssl} = data;
   try {
-    const res = await request(`http${ssl ? 's' : ''}://${host}:${port}/wd/hub/sessions`);
+    const res = await request(`http${ssl ? 's' : ''}://${host}:${port}${path ? path : '/wd/hub'}/sessions`);
     evt.sender.send('appium-client-get-sessions-response', {res});
   } catch (e) {
     evt.sender.send('appium-client-get-sessions-fail');
--- a/app/renderer/actions/Session.js
+++ b/app/renderer/actions/Session.js
@@ -552,14 +552,14 @@ export function setSavedServerParams () {
 
 export function getRunningSessions () {
   return (dispatch, getState) => {
-    const avoidGetRunningSessions = ['sauce', 'testobject'];
+    const avoidServerTypes = ['sauce', 'testobject'];
     // Get currently running sessions for this server
     const state = getState().session;
     const {server, serverType} = state;
     const serverInfo = server[serverType];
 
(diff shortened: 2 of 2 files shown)


app/main/appium.js
```diff
@@ -321,9 +321,9 @@ function connectClientMethodListener () {
 }
 
 const getCurrentSessions = _.debounce(async (evt, data) => {
-  const {host, port, ssl} = data;
+  const {host, port, path, ssl} = data;
```
mykola-mokhnach: this is overriding the default import, which is also named `path`. Would it be possible to rename the prop or just use `data.path` format to avoid shadowing?

dpgraham: I would do something like....

`const {host, port, path:appiumPath, ssl} = data;`

app/main/appium.js
```diff
@@ -321,9 +321,9 @@ function connectClientMethodListener () {
...
   try {
-    const res = await request(`http${ssl ? 's' : ''}://${host}:${port}/wd/hub/sessions`);
+    const res = await request(`http${ssl ? 's' : ''}://${host}:${port}${path ? path : '/wd/hub'}/sessions`);
```
dpgraham: You can just set 'wd/hub' as the default above

`const {host, port, path:appiumPath = 'wd/hub', ssl} = data;`

And then this is just:

`http${ssl ? 's' : ''}://${host}:${port}${appiumPath}/sessions`

app/renderer/actions/Session.js
```diff
@@ -214,10 +214,10 @@ export function newSession (caps, attachSessId = null) {
...
       case ServerTypes.headspin: {
         const headspinUrl = url.parse(session.server.headspin.webDriverUrl);
-        host = headspinUrl.hostname;
-        port = headspinUrl.port;
-        path = headspinUrl.pathname;
-        https = headspinUrl.protocol === 'https:';
+        host = session.server.headspin.hostname = headspinUrl.hostname;
+        port = session.server.headspin.port = headspinUrl.port;
+        path = session.server.headspin.path = headspinUrl.pathname;
+        https = session.server.headspin.ssl = headspinUrl.protocol === 'https:';
```
dpgraham: Do us other cloud providers need to set path too?

app/renderer/actions/Session.js
```diff
@@ -552,14 +552,19 @@ export function setSavedServerParams () {
 
 export function getRunningSessions () {
   return (dispatch, getState) => {
+    const avoidServerTypes = ['sauce', 'testobject'];
```
dpgraham: Why is it only Sauce being excluded in this?

KazuCocoa: I didn't change this logic. (Just switched to use `includes` method.)
Ordinary, you implemented this logic. https://github.com/appium/appium-desktop/pull/196
Can I remove here?

app/renderer/actions/Session.js
KazuCocoa: I'll take a look at other vendors. 👍 
I wondered the path was intentional.

I'll ping proper persons if others also need to consider here.
(And change as another PRs)

app/main/appium.js
KazuCocoa: oops... nice..

app/renderer/actions/Session.js
KazuCocoa: https://github.com/appium/appium-desktop/compare/master...KazuCocoa:update-session-server-vendors?expand=1
I'll work for it after this PR

app/renderer/actions/Session.js
dpgraham: Ok cool. That's fine then.

app/renderer/actions/Session.js
dpgraham: @KazuCocoa we can let them take care of it themselves. I just wanted to make sure it wasn't breaking anything, which I don't think it will.

dpgraham:
lgtm

app/renderer/actions/Session.js
KazuCocoa: 👍

Yeah, I'm thinking to create the PR and mention them if they would like to follow it

PR #983, opened 2019-05-13
apply session.server.${providername}.host/port/path/ssl

KazuCocoa:
https://github.com/appium/appium-desktop/pull/979 for other providers.

AppiumDesktop has a feature which gets current running sessions by Appium Server via `getSessions` command like below image. (POST to `https://host:port/wd/hub/sessions`) Before https://github.com/appium/appium-desktop/pull/979 , it worked partially.

image

Make it work correctly, below values are necessary.
I've added them as this PR.

```
session.server.${providername}.host
session.server.${providername}.port
session.server.${providername}.path
session.server.${providername}.ssl # true or false
```

I assume some of providers won't work `getSessions` command on your environment.
Then, you can ignore the feature like adding your name like below.
https://github.com/appium/appium-desktop/blob/44c7f49ce482e25fca3c86f6a353bed1963f2994/app/renderer/actions/Session.js#L555

Could you check this PR and add your name in https://github.com/appium/appium-desktop/blob/44c7f49ce482e25fca3c86f6a353bed1963f2994/app/renderer/actions/Session.js#L555 if you would like to disable the feature for your environment?

@UziEilon @mohitmun @Walther @khanhdodang @shibupanda @testingbot
(I mentioned you who created PRs to add each provider's tab)

I'm thinking to merge this after getting approval and over a week later.

### DIFF
--- a/app/renderer/actions/Session.js
+++ b/app/renderer/actions/Session.js
@@ -552,7 +552,9 @@ export function setSavedServerParams () {
 
 export function getRunningSessions () {
   return (dispatch, getState) => {
-    const avoidServerTypes = ['sauce', 'testobject'];
+    const avoidServerTypes = [
+      'sauce', 'testobject', 'testingbot'
+    ];
     // Get currently running sessions for this server
     const state = getState().session;
     const {server, serverType} = state;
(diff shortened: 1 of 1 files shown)


KazuCocoa:
Thanks for the response. I'll do in this PR.

PR #1005, opened 2019-06-12
Add languages to interface

dpgraham:
@jlipps let me know when we merge this then I'll make a new release.


assets/locales/hi/translation.json
```diff
@@ -0,0 +1,319 @@
+{
+  "en": "English",
+  "de": "Deutsch",
+  "ru": "रूसी",
```
mykola-mokhnach: language names should be translated equal for all locales

mykola-mokhnach:
The new language names should also be added to the main translation resource

mykola-mokhnach:
BTW, the tests seem to be flaky

dpgraham:
@mykola-mokhnach the 'tags' build always fails because it tries publishing the same Windows version twice. I just need to take that out.

I'll ask the translators to do the locale translation.

PR #1014, opened 2019-06-20
Push Crowdin changes directly to master

dpgraham:
The weekly pipeline that updates the crowdin translations and the package-lock.json has been routinely approved and merged for months now. I think it's safe to just push directly to master and skip the pull requests.


mykola-mokhnach:
The crowdin repository is opened for everyone, which means one can add "malicious" translations and this will be "blindly" merged. Do we have any solution for that?

dpgraham:
@mykola-mokhnach the changes would be pushed but not released.

PR #1030, opened 2019-07-04
WIP: Testing still... Revert "Get source, screenshot and windowSize concurrently"

dpgraham:
Reverts appium/appium-desktop#916

Causing race conditions that cause the spinner to act indefinitely

### DIFF
--- a/app/main/appium-method-handler.js
+++ b/app/main/appium-method-handler.js
@@ -166,37 +166,36 @@ export default class AppiumMethodHandler {
   }
 
   async _getSourceAndScreenshot () {
+    let source, sourceError, screenshot, screenshotError, windowSize, windowSizeError;
+    try {
+      source = await this.driver.source();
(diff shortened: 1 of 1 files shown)


dpgraham:
This doesn't fix the issue.

vijayqaguy:
@dpgraham I am just providing more information to narrow down this issue. It works perfectly fine on v1.12.0 but not on v1.12.1. We just need to compare the changes done between these versions. I am trying from my side as well.

danieltg:
@dpgraham I've just tried it now and Perfecto cloud provider run properly after reverting commit #916 back.

danieltg:
@dpgraham Can you kindly retry the revert as per our testing it does fix the issue - at least for Perfecto cloud provider.

dpgraham:
@danieltg I don't have time, but I'll just leave this as-is until I get around to it again

PR #1036, opened 2019-07-09
fix: Prevent wd from injecting default desired capabilities

### LINKED ISSUE #667: Extra browser capability appears and session cannot be created
## Appium or Appium Desktop?
Appium Desktop

## The problem

I'm trying to start appium desktop for an android device using capabilities and appium desktop:
```
{
"platformName": "android",
"platformVersion": "7",
"deviceName": "huawei",
"app": "C:\automation\yamba-debug.apk",
"automationName": "uiautomator2",
"udid": "a350b460"
}
```
and i get
[BaseDriver] SessionNotCreatedError: A new session could not be created. Details: The desiredCapabilities object was not valid for the following reason(s): 'deviceName' can't be blank

In logs i can see that some capabilities are put by default, especially:
[Appium] browserName: firefox
which can be the source of the problem.

Did something changed in latest version? 

Logs:
https://gist.github.com/vjmaiky/0e2109d830d2fbb0f9eaf4251500b3d9

## Environment
Windows x64, Appium desktop 1.7.1, Appium server 1.9.1, Real Android device, Android 7 and 6

- I am running Appium Desktop version _<1.7.1>_.
- I am on (pick one):
    - [ ] Mac
    - [x ] Windows
    - [ ] Linux

Volune:
If the capability `deviceName` is not defined (for example if it is defined by appium server's default capabilities), this PR will prevent  `wd` from injecting `"browserName": "firefox"`.

See https://github.com/admc/wd/blob/13ffe77a1a599dcff5dcb8242c003e14cfcab57d/lib/webdriver.js#L83-L91

Fixes #667

### DIFF
--- a/app/main/appium.js
+++ b/app/main/appium.js
@@ -209,6 +209,12 @@ function connectCreateNewSession () {
         desiredCapabilities.connectHardwareKeyboard = true;
       }
 
+      // Prevent wd from injecting default desired capabilities
+      if (typeof desiredCapabilities.wdNoDefaults === 'undefined' &&
+          typeof desiredCapabilities['wd-no-defaults'] === 'undefined') {
+        desiredCapabilities.wdNoDefaults = true;
+      }
+
       // Try initializing it. If it fails, kill it and send error message to sender
       let p = driver.init(desiredCapabilities);
       event.sender.send('appium-new-session-successful');


dpgraham:
Thank you!

PR #1072, opened 2019-08-16
feat: Remove testingbot from avoidServerTypes in getRunningSessions

testingbot:
Hi,

We've made sure that `getRunningSessions` will work, please remove testingbot from the avoidServerTypes.

Thanks


KazuCocoa:
thanks

PR #1081, opened 2019-09-08
docs: add how to connect to local appium server section

KazuCocoa:
Recently I guided to connect to local appium server in order to use `appium@beta` or workaround.
Then, I frequently explained how to do it.

I think it's useful to have a guide to connect to local appium server, but outside appium desktop in readme. Thus, this PR has it in the troubleshooting section.


dpgraham:
Thank you so much for this! I've wanted to add this for a long time.

PR #1102, opened 2019-10-01
fix: Revert electron-builder in order to complete publishing package tasks

KazuCocoa:
I guess https://github.com/appium/appium-desktop/pull/1047/files is the cause about the failing tasks.

https://dev.azure.com/AppiumCI/Appium%20CI/_build/results?buildId=5650

```
[command]/bin/bash --noprofile --norc /Users/vsts/agent/2.155.1/work/_temp/7b90ec96-5cc2-4f6b-9763-b5a07a5adf28.sh
  • electron-builder  version=21.2.0 os=18.7.0
  • loaded configuration  file=package.json ("build" field)
  • packaging       platform=darwin arch=x64 electron=5.0.6 appOutDir=release/mac
  • asar using is disabled — it is strongly not recommended  solution=enable asar and use asarUnpack to unpack files that must be externally available
  ⨯ /Users/vsts/agent/2.155.1/work/1/s/$(CSC_LINK) doesn't exist
```

Then, maybe https://github.com/appium/appium-desktop/pull/1079 also should be reverted.

----

electron-builder 21.0 has a breaking change. https://github.com/electron-userland/electron-builder/releases/tag/v21.0.1

We need to take a look more when we update it. (Maybe after releasing 1.15.0)

@dpgraham What do you think to try to run release task with this PR once, and update electron-builder again to investigate the cause.

### DIFF
--- a/ci-jobs/templates/package.yml
+++ b/ci-jobs/templates/package.yml
@@ -7,7 +7,7 @@ parameters:
   pool: ''
   xvfb: false
   target: ''
-  buildScript: 'npx electron-builder build --publish always'
+  buildScript: 'npx build --publish always'
 jobs:
   - job: ${{parameters.name}}
     pool: ${{parameters.pool}}
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,9 +11,9 @@
       "dev": true
     },
     "7zip-bin": {
-      "version": "5.0.3",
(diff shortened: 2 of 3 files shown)


KazuCocoa:
ah, manual start to build release packages?
https://dev.azure.com/AppiumCI/Appium%20CI/_build?definitionId=41&_a=summary

KazuCocoa:
oops... the same error is happening... mm
https://github.com/appium/appium-desktop/runs/243542526

PR #1106, opened 2019-10-02
fix: Revert "fix: Revert electron-builder in order to complete publishing package tasks"

KazuCocoa:
Reverts appium/appium-desktop#1102 since this revert did not work...

https://github.com/appium/appium-desktop/runs/243542526


KazuCocoa:
ah.... the errors seem below environment variables...

https://dev.azure.com/AppiumCI/Appium%20CI/_build/results?buildId=5709

Could you take a look they are set when you have a chance? @dpgraham 

https://github.com/appium/appium-desktop/blob/cadf4403a43e35b28e619121997304eae66688bd/ci-jobs/templates/package.yml#L31-L33

KazuCocoa:
Thanks!

image

They were drafted. We should make them release. 👀 
hm.
Will take a look a bit when I have time.

PR #1137, opened 2019-10-19
fix: shows newer version available dialog only when the local version is lower than it

KazuCocoa:
`res` gets value if the latest release tagged version on github is not the current local app version.
Thus, update dialog appears when a user has a newer pre-released version than the latest release tagged version on github.

e.g.
A user has 1.15.0 on their local, but this repository's release tagged version is 1.14.0
=> Then, the local version shows update dialog to 1.14.0

We can see questions about this issue on this repository. So, I've added version comparison after getting the `res`. So, if the local version is greater than the github one, no available device dialog appears.

### DIFF
--- a/app/main/auto-updater/update-checker.js
+++ b/app/main/auto-updater/update-checker.js
@@ -1,10 +1,11 @@
 import request from 'request-promise';
 import { getFeedUrl } from './config';
+import semver from 'semver';
 
 export async function checkUpdate (currentVersion) {
   try {
     const res = await request.get(getFeedUrl(currentVersion));
-    if (res) {
+    if (res && semver.lt(currentVersion, res.name)) {
       return JSON.parse(res);
     }
   } catch (ign) { }
--- a/app/main/auto-updater/update-checker.js
+++ b/app/main/auto-updater/update-checker.js
@@ -4,6 +4,11 @@ import semver from 'semver';
 
 export async function checkUpdate (currentVersion) {
   try {
+    // The response is like (macOS):
+    // {  "name":"v1.15.0-1",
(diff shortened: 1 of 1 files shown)


app/main/auto-updater/update-checker.js
```diff
@@ -1,11 +1,15 @@
 import request from 'request-promise';
 import { getFeedUrl } from './config';
+import semver from 'semver';
 
 export async function checkUpdate (currentVersion) {
   try {
     const res = await request.get(getFeedUrl(currentVersion));
     if (res) {
-      return JSON.parse(res);
+      const j = JSON.parse(res);
```
mykola-mokhnach: `const {name} = `

app/main/auto-updater/update-checker.js
```diff
@@ -1,11 +1,15 @@
 import request from 'request-promise';
 import { getFeedUrl } from './config';
+import semver from 'semver';
 
 export async function checkUpdate (currentVersion) {
   try {
     const res = await request.get(getFeedUrl(currentVersion));
```
mykola-mokhnach: how the remote response look like?

KazuCocoa: like: `{"name":"v1.15.0-1","notes":"* Bump up Appium to v1.15.0","pub_date":"2019-10-04T04:40:37Z","url":"https://github.com/appium/appium-desktop/releases/download/v1.15.0-1/Appium-1.15.0-1-mac.zip"}` (mac)
The `name` is tag name.

mykola-mokhnach: it could be useful to have it mentioned  in comments

Issue #957, opened 2019-04-27
Element properties don't wrap up on the view


Juanap12:
## Appium or Appium Desktop?

You are reporting an issue at the Appium Desktop repository. Appium Desktop is a wrapper around Appium. If you are having trouble running tests, it is much more likely that the problem you are encountering is not a problem with Appium Desktop but with Appium. For that reason we require you to have tried your testcase with Appium CLI before reporting issues. Check the checkbox below to confirm that you have proven your issue does _not_ reproduce on Appium itself:

- [x] I have verified that my issue does not occur with Appium and should be investigated as an Appium Desktop issue

## The problem

_Whenever I inspect an element, the description or value of an elements property goes beyond the screen. Even if I make the screen bigger, the container doesn’t wraps the text._

![Screen Shot 2019-04-26 at 3 43 47 PM](https://user-images.githubusercontent.com/30246409/56850453-97bcee00-68c8-11e9-83c5-34a2c3c509b0.jpg)

## Environment

- I am running Appium Desktop version _<1.12.1>_.
- I am on (pick one):
    - [x] Mac
    - [ ] Windows
    - [ ] Linux

PR #1170, opened 2019-11-10
fix: break word in selected element

KazuCocoa:
- before

- after

----

closes https://github.com/appium/appium-desktop/issues/957 

This trick was linked in https://ant.design/components/table/

### DIFF
--- a/app/renderer/components/Inspector/SelectedElement.js
+++ b/app/renderer/components/Inspector/SelectedElement.js
@@ -8,8 +8,6 @@ import { clipboard } from 'electron';
 
 const ButtonGroup = Button.Group;
 
-const tableColumnsStyle = { wordWrap: 'break-all', wordBreak: 'break-all' };
-
 /**
  * Shows details of the currently selected element and shows methods that can
  * be called on the elements (tap, sendKeys)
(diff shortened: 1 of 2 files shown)


app/renderer/components/Inspector/SelectedElement.js
```diff
@@ -8,6 +8,8 @@ import { clipboard } from 'electron';
 
 const ButtonGroup = Button.Group;
 
+const tableColumnsStyle = { wordWrap: 'break-all', wordBreak: 'break-all' };
```
mykola-mokhnach: cannot we define the style in css file?

KazuCocoa: Could 🆗 
Maybe something was wrong when I tried once...

app/renderer/components/Inspector/SelectedElement.js
```diff
@@ -46,11 +46,14 @@ class SelectedElement extends Component {
       title: t('Attribute'),
       dataIndex: 'name',
       key: 'name',
-      width: 100
+      width: 100,
+      render: (text) => (<div className={styles['selected-element-table-cells']}>{text}</div>)
```
mykola-mokhnach: how about extracting the cell markup function to avoid unnecessary duplication?

KazuCocoa:
hm, the failing test was probably timing... (The attribute exists in the view as below)
image

PR #1171, opened 2019-11-10
fix: add scroll in saved capability sets

KazuCocoa:
- before: We cannot get items in a hidden area. The below image had 11 items, but we can select 3 of 11. They appear after resizing the size of the window.

image

- after:

image

We should calculate the available area dynamically

### DIFF
--- a/app/renderer/components/Session/SavedSessions.js
+++ b/app/renderer/components/Session/SavedSessions.js
@@ -5,7 +5,7 @@ import { Button, Row, Col, Table } from 'antd';
 import FormattedCaps from './FormattedCaps';
 import SessionCSS from './Session.css';
 
-const heightOfServerConfigArea = 400;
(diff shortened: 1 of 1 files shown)


app/renderer/components/Session/SavedSessions.js
```diff
@@ -1,13 +1,14 @@
 import React, { Component } from 'react';
 import moment from 'moment';
+import { remote } from 'electron';
 import { Button, Row, Col, Table } from 'antd';
 import FormattedCaps from './FormattedCaps';
 import SessionCSS from './Session.css';
 
+const heightOfServerConfigArea = 400;
```
mykola-mokhnach: usually constant names are in UPPERCASE

KazuCocoa:
cc @dpgraham

Issue #1173, opened 2019-11-11
Appium Error - An unknown server-side error occurred while processing the command. Original error: Callback must be a function


testmonk:
## The problem
Receiving the below error while launching the iOS app.
An unknown server-side error occurred while processing the command. Original error: Callback must be a function

##Environment
Mac

- I am running Appium Desktop version _<1.15.1 (1.15.1.20191013.2)>_.

@KazuCocoa Please advice.

KazuCocoa:
Could you share the appium log?

testmonk:
@KazuCocoa Check this out.

[info] [35m[Appium][39m Welcome to Appium v1.15.1
[info] [35m[Appium][39m Non-default server args:
[info] [35m[Appium][39m   allowInsecure: {
[info] [35m[Appium][39m   }
[info] [35m[Appium][39m   denyInsecure: {
[info] [35m[Appium][39m   }
[info] [35m[Appium][39m Appium REST http interface listener started on 0.0.0.0:4723[info] [35m[HTTP][39m [37m-->[39m [37mGET[39m [37m/wd/hub/sessions[39m
[info] [35m[HTTP][39m [90m{}[39m
[debug] [35m[GENERIC][39m Calling AppiumDriver.getSessions() with args: []
[debug] [35m[GENERIC][39m Responding to client with driver.getSessions() result: []
[info] [35m[HTTP][39m [37m<-- GET /wd/hub/sessions [39m[32m200[39m [90m6 ms - 40[39m
[info] [35m[HTTP][39m [90m[39m[info] [35m[HTTP][39m [37m-->[39m [37mPOST[39m [37m/wd/hub/session[39m
[info] [35m[HTTP][39m [90m{"desiredCapabilities":{"agentPath":"/usr/local/lib/node_modules/appium/node_modules/appium-xcuitest-driver/WebDriverAgent/WebDriverAgent.xcodeproj","app":"/Users/admin/Downloads/mho.ipa","automationName":"XCUITest","bootstrapPath":"/usr/local/lib/node_modules/appium/node_modules/appium-xcuitest-driver/WebDriverAgent","bundleId":"com.******.devteam","deviceName":"YIP-51","platformName":"iOS","platformVersion":"12.3.1","udid":"00008020-001A65D02188002E","useNewWDA":true,"xcodeOrgId":"849UP4BSN8","xcodeSigningId":"iPhone Developer","newCommandTimeout":0,"connectHardwareKeyboard":true}}[39m
[...]

PR #1182, opened 2019-11-17
feat: add running env of nodejs, electron

KazuCocoa:
Running NodeJS version will help us/users to know what nodejs version is used in AD in about app menu.
e.g. https://github.com/appium/appium-desktop/issues/1173

image


app/main/menus.js
```diff
@@ -7,6 +7,11 @@ import i18n from '../configs/i18next.config';
 let menuTemplates = {mac: {}, other: {}};
 let mainWindow = null;
 
+const appVersionTitle = i18n.t('Appium Desktop');
+const appVersionMessage = `App Version: ${app.getVersion()}\n` +
```
mykola-mokhnach: `App Version` should also be translated.
I would recommend putting the whole string into resources though, because in RTL languages it might be spelled differently even if the particular words don't need to be translated

app/main/menus.js
```diff
@@ -195,8 +206,8 @@ function otherMenuFile () {
     label: i18n.t('&About Appium'),
     click () {
       dialog.showMessageBox({
-        title: i18n.t('Appium Desktop'),
-        message: i18n.t(`Version ${app.getVersion()}`),
```
mykola-mokhnach: Please remove the string from resources if it is not needed anymore

app/main/menus.js
KazuCocoa: ah, RTL... Forgot it

app/main/menus.js
```diff
@@ -29,8 +24,12 @@ function macMenuAppium () {
       label: i18n.t('About Appium'),
       click () {
         dialog.showMessageBox({
```
mykola-mokhnach: I'd rather extract this dialog into separate module to avoid unnecessary duplication

app/main/menus.js
```diff
@@ -203,14 +205,7 @@ function otherMenuFile () {
   }, {
     label: i18n.t('&About Appium'),
     click () {
```
mykola-mokhnach: `click: showAppInfoClickAction,` should also work fine

KazuCocoa: ah... I tried `click: showAppInfoClickAction()` but it called at launching time, but `click: showAppInfoClickAction` worked as expected. 🤯

PR #1188, opened 2019-11-24
feat: update electron version to v6

KazuCocoa:
Update electron version to v6.
v7 is available, but let me update the version one by one.

- `componentWillMount` was changed.
    - Move the logic which has fetching data into `didMount`
    ```
    /Users/kazu/GitHub/appium-desktop/node_modules/react-dom/cjs/react-dom.development.js:11494 Warning: componentWillMount has been renamed, and is not recommended for use. See https://fb.me/react-async-component-lifecycle-hooks for details.

    * Move code with side effects to componentDidMount, and set initial state in the constructor.
    * Rename componentWillMount to UNSAFE_componentWillMount to suppress this warning in non-strict mode. In React 17.x, only the UNSAFE_ name will work. To rename all deprecated lifecycles to their new names, you can run `npx react-codemod rename-unsafe-lifecycles` in your project source folder.
    ```
    - I checked each component can load data by maual
- `componentWillUpdate`
    - https://reactjs.org/docs/react-component.html#getsnapshotbeforeupdate
    - `ServerMonitor.js` will scroll down automatically if some new logs come.
        - I checked the behaviour with the current change.
        - When there was no `return null`, an error that required to return something (not undefined) happened. The return value was used in `componentDidUpdate`, but we did not use it. so, return `null` was enough to prevent the error.

### DIFF
--- a/app/renderer/components/ServerMonitor/ServerMonitor.js
+++ b/app/renderer/components/ServerMonitor/ServerMonitor.js
@@ -121,6 +121,7 @@ export default class ServerMonitor extends Component {
     if (n) {
       this.shouldScroll = n.scrollTop + n.offsetHeight >= n.scrollHeight;
     }
+    return null;
   }
 
   componentDidUpdate () {
--- a/app/renderer/components/ServerMonitor/ServerMonitor.js
+++ b/app/renderer/components/ServerMonitor/ServerMonitor.js
@@ -115,7 +115,7 @@ export default class ServerMonitor extends Component {
     document.removeEventListener('keydown', this.keydownListener);
   }
 
-  componentWillUpdate () {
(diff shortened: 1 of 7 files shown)


app/renderer/components/ServerMonitor/ServerMonitor.js
```diff
@@ -115,12 +115,13 @@ export default class ServerMonitor extends Component {
     document.removeEventListener('keydown', this.keydownListener);
   }
 
-  componentWillUpdate () {
+  getSnapshotBeforeUpdate () {
```
mykola-mokhnach: I don't see the connection between the name of this method and what it is actually doing
Perhaps, we could set a better name to it?

PR #1231, opened 2020-01-26
feat: Add context and some web commands as actions

KazuCocoa:
Added some commands for web/context

`methodName` were in https://github.com/admc/wd/blob/67dae3823e8a457c10d9fbe81578684bbb9938d9/doc/api.md
Command tree and their name were http://appium.io/docs/en/commands

I've tested they worked on my local


KazuCocoa:
@dpgraham Do you have a chance to take a glance?

Issue #1254, opened 2020-03-01
UI components have become unscrollable for Electron > 3


dpgraham:
## The problem

The "Saved Sessions" and the JSON representation of the selected session became unscrollable when upgrading from Electron v3 ~> Electron v4 (I've confirmed this in my tests)

Here's a video:

https://www.dropbox.com/s/oelmt6auu3ad3ff/Screen%20Recording%202020-02-29%20at%207.56.24%20PM.mov?dl=0

Observations:
* Since it's been isolated to the Electron version and it's a UI problem, it's likely that something in a later Chromium version broke the UI
* The likely fix is to just tweak the CSS

## Environment

Happens on Mac

PR #1255, opened 2020-03-01
fix: can scroll in FormattedCaps and NewSessionForm

KazuCocoa:
Fix https://github.com/appium/appium-desktop/issues/1254

https://www.dropbox.com/s/hxxxwh5b1ehpl63/scroll.mov?dl=0

@dpgraham What about like this?


app/renderer/components/Session/Session.js
```diff
@@ -16,6 +16,8 @@ const {TabPane} = Tabs;
 
 const ADD_CLOUD_PROVIDER = 'addCloudProvider';
 
+const HEIGHT_OF_SERVICE_CONFIG_AREA = 400;
```
mykola-mokhnach: I'd rather define the constant in a separate file to avoid possible cross references

app/renderer/components/Session/NewSessionForm.js
```diff
@@ -3,9 +3,20 @@ import { Button, Input, Modal, Form, Row, Col, Select } from 'antd';
...
 import CapabilityControl from './CapabilityControl';
 import SessionStyles from './Session.css';
+import {remote} from 'electron';
+import { HEIGHT_OF_SERVICE_CONFIG_AREA } from './Session.js';
+
 const {Item: FormItem} = Form;
 const {Option} = Select;
 
+const styles = {
+  // Got warning messages not to set by CSS
```
mykola-mokhnach: I'd also use CSS to define styles. I'm surprised that electron prints such warning

mykola-mokhnach:
Is there any chance to fix failing test on Windows or they are unrelated?

dpgraham:
I'm going to re-run the Windows test. It's a UI test failure though, so I'd like to confirm if it's a bug or not before merging this.

dpgraham:
Looks like a race condition. Probably means the UI is fine in Windows but we should still test it out before next release.

app/renderer/components/Session/NewSessionForm.js
KazuCocoa: I found a wrong syntax in my previous CSS implementation

app/renderer/components/Session/Layout.js
```diff
@@ -0,0 +1,3 @@
+const HEIGHT_OF_SESSION_CONFIG_AREA = 400;
```
mykola-mokhnach: 👍

PR #1259, opened 2020-03-08
feat: Update to antd v4

KazuCocoa:
Update `antd` to v4.
The migration guide is https://ant.design/docs/react/migration-v4

The main change is `Icon`.

I tested on my local.

- Opened All windows
- Did actions we can conduct via icons


package.json
```diff
@@ -136,7 +136,7 @@
     "moment": "^2.20.1",
     "postcss": "^7.0.4",
     "prop-types": "^15.6.2",
-    "react": "^16.9.0",
+    "react": "^16.13.0",
```
KazuCocoa: antd v4 requires over 16.12

app/renderer/components/Inspector/RecordedActions.js
```diff
@@ -89,7 +101,7 @@ class RecordedActions extends Component {
 
     const highlightedCode = this.code(false);
 
-    return <Card title={<span><Icon type="code-o"/> {t('Recorder')}</span>}
+    return <Card title={<span><LegacyIcon type="code-o"/> {t('Recorder')}</span>}
```
mykola-mokhnach: in antd documentation it is stated that `Legacy Icon usage will be discarded`. Should we already stop using it?

KazuCocoa: Yes. I'm thinking to work for the LegacyIcon stuff as following PRs since I need to dig them in more

PR #1302, opened 2020-03-23
fix: Use treeData prop for Tree component instead of deprecated children prop

rhysd:
I noticed following warning appeared in DevTools console:

キャプチャ2

It seems that `treeData` property should be used to set tree nodes data.

https://ant.design/components/tree/

This PR fixes the point. After applying this patch, I confirmed the warning no longer occurred and tree view worked fine by connecting to sample Android app as follows:

キャプチャ

### DIFF
--- a/app/renderer/components/Inspector/Source.js
+++ b/app/renderer/components/Inspector/Source.js
@@ -4,7 +4,6 @@ import LocatorTestModal from './LocatorTestModal';
 import InspectorStyles from './Inspector.css';
 import { withTranslation } from '../../util';
 
-const {TreeNode} = Tree;
 const IMPORTANT_ATTRS = [
   'name',
   'content-desc',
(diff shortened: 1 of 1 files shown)


app/renderer/components/Inspector/Source.js
```diff
@@ -72,9 +71,11 @@ class Source extends Component {
       if (elemObj.children.length === 0) {return null;}
 
       return elemObj.children.map((el) => {
-        return <TreeNode title={this.getFormattedTag(el)} key={el.path}>
-          {recursive(el)}
-        </TreeNode>;
+        return {
```
mykola-mokhnach: this could be changed to 
```
return elemObj.children.map((el) => ({
 title: this.getFormattedTag(el),
 key: el.path,
 children: recursive(el),
}));
```

PR #1303, opened 2020-03-23
fix: set Perfecto cloud provider default port when undefined

Dan-Maor:
Sometimes the cloud port is not sent correctly from the UI, therefor setting it to the default in Session.js when it is undefined.

### DIFF
--- a/app/renderer/actions/Session.js
+++ b/app/renderer/actions/Session.js
@@ -222,7 +222,7 @@ export function newSession (caps, attachSessId = null) {
       }
       case ServerTypes.perfecto:
         host = session.server.perfecto.hostname;
-        port = session.server.perfecto.port;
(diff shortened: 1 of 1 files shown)


app/renderer/actions/Session.js
```diff
@@ -222,7 +222,7 @@ export function newSession (caps, attachSessId = null) {
       }
       case ServerTypes.perfecto:
         host = session.server.perfecto.hostname;
-        port = session.server.perfecto.port;
+        port = session.server.perfecto.port || 80;
```
mykola-mokhnach: should it be `?? 80` or zero value is also invalid?

Dan-Maor: 0 should also be invalid in this case.

PR #1321, opened 2020-04-06
Fix 'Saved Session' table is not clickable

rhysd:
I noticed saved session was not clickable. In table of 'Saved Sessions', `onClick` is used for handling click events. However, as far as I checked it seemed mistake. There is no `onClick` property.

https://ant.design/components/table/#Table

Instead, `onRow` property is available.

https://ant.design/components/table/#onRow-usage

After applying this PR, I confirmed click got effective.


KazuCocoa:
thanks!

PR #1330, opened 2020-04-08
fix: warnings for type

KazuCocoa:
Fixes warnings about `type`. Their default should be `default`. `default` is used in some places, but not rest of them are still `null` or `''` (blank)
https://ant.design/components/button/

They seem okay, but let me check them again later

### DIFF
--- a/app/renderer/components/Inspector/RecordedActions.js
+++ b/app/renderer/components/Inspector/RecordedActions.js
@@ -50,8 +50,6 @@ class RecordedActions extends Component {
       {frameworks[f].readableName}
     </Option>);
 
-    let boilerplateType = showBoilerplate ? 'primary' : 'default';
-
     return <div>
       {!!recordedActions.length &&
         <Select defaultValue={actionFramework} onChange={setActionFramework}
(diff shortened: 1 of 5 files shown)


app/renderer/components/Inspector/RecordedActions.js
```diff
@@ -66,7 +64,7 @@ class RecordedActions extends Component {
             <Button
               onClick={toggleShowBoilerplate}
               icon={<ExportOutlined/>}
-              type={boilerplateType}
+              type={showBoilerplate ? 'primary' : 'default'}
```
mykola-mokhnach, line 67: should these magic strings be moved into constants?

KazuCocoa:
I checked all pages I switched with 'default' from null

app/renderer/components/Inspector/RecordedActions.js
```diff
@@ -66,7 +65,7 @@ class RecordedActions extends Component {
             <Button
               onClick={toggleShowBoilerplate}
               icon={<ExportOutlined/>}
-              type={boilerplateType}
+              type={showBoilerplate ? BUTTON_PRIMARY : BUTTON_DEFAULT}
```
mykola-mokhnach, line 68: we could still reduce the duplication by having that logic in a function, for example:
```
function getButtonType (showBoilerplate) {
    return showBoilerplate ? BUTTON_PRIMARY : BUTTON_DEFAULT;
}
```

PR #1336, opened 2020-04-11
chore: Replace strings with const for antd elements

KazuCocoa:
Replace string with const for `type` of antd elements


app/renderer/components/Inspector/Actions.js
```diff
@@ -104,7 +105,14 @@ export default class Actions extends Component {
...
-              {argType === NUMBER && <Input type="number" value={pendingAction.args[index]} addonBefore={t(argName)} onChange={(e) => setActionArg(index, _.toNumber(e.target.value))} />}
+              {
+                argType === NUMBER && <Input
+                  type={INPUT.NUMBER}
+                  value={pendingAction.args[index]}
+                  addonBefore={t(argName)}
+                  onChange={(e) => setActionArg(index, _.toNumber(e.target.value))}
+                />
+              }
               {argType === STRING && <Input addonBefore={t(argName)} onChange={(e) => setActionArg(index, e.target.value)}/>}
```
mykola-mokhnach, line 116: I was thinking about adding a linter rule to enforce each property to be on a new line. What do you think @KazuCocoa ?

app/renderer/components/AntdType.js
```diff
@@ -1,6 +1,32 @@
-// Button
```
mykola-mokhnach, line 1: `AntdTypes` would be a more precise name for the file

app/renderer/components/Inspector/Actions.js
KazuCocoa: I also thought if we could coordinate syntax by lint

PR #1337, opened 2020-04-12
fix: warnings in Actions

KazuCocoa:
This change fixes two warnings in Actions.

- React.createElement: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: object. antd select
- a unique key

### DIFF
--- a/app/renderer/components/Inspector/Actions.js
+++ b/app/renderer/components/Inspector/Actions.js
@@ -5,10 +5,6 @@ import { actionDefinitions, actionArgTypes } from './shared';
 import InspectorStyles from './Inspector.css';
 import { INPUT } from '../AntdTypes';
 
-const { STRING, NUMBER } = actionArgTypes;
-
-const Option = { Select };
-
 export default class Actions extends Component {
 
   startPerformingAction (actionName, action) {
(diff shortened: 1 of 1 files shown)


app/renderer/components/Inspector/Actions.js
```diff
@@ -78,19 +74,19 @@ export default class Actions extends Component {
...
-            { _.keys(actionDefinitions).map((actionGroup) => <Option key={actionGroup}>{t(actionGroup)}</Option>) }
+            { _.keys(actionDefinitions).map((actionGroup) => <Select.Option key={actionGroup}>{t(actionGroup)}</Select.Option>) }
           </Select>
         </Col>
       </Row>
       {selectedActionGroup && <Row>
         <Col span={24}>
           <Select onChange={(actionGroupName) => selectSubActionGroup(actionGroupName)} placeholder={t('Select Sub Group')}>
-            { _.keys(actionDefinitions[selectedActionGroup]).map((actionGroup) => <Option key={actionGroup}>{t(actionGroup)}</Option>) }
+            { _.keys(actionDefinitions[selectedActionGroup]).map((actionGroup) => <Select.Option key={actionGroup}>{t(actionGroup)}</Select.Option>) }
```
mykola-mokhnach, line 84: Some lines here are really long

KazuCocoa: Yeah, would like to do lint stuff as another pr

PR #1372, opened 2020-04-30
chore: update bot cycle weekly

KazuCocoa:
dependabot/config.yml does not support format like ignoring only `patch`, but we want to reduce PR frequency.
Let me change the update schedule to weekly for now


KazuCocoa:
will merge this soon

Issue #1410, opened 2020-06-03
Could not start session, when starting appium inspector 


Amrkamel1:
## Appium or Appium Desktop?

Unable to start inspector session for browser stack 
i had to downgrade again to .17.0 to make it work

## The problem

unable to start appium inspector due to the following error 

![appium](https://user-images.githubusercontent.com/32042806/83665088-443cc980-a5cb-11ea-82f3-ca0aecca0012.png)

## Environment

- I am running Appium Desktop version _<1.17.1>_.
- I am on (pick one):
    - [ ] Linux

KazuCocoa:
> i had to downgrade again to .17.0 to make it work

Is this Appium Desktop version?
I thought something error returned from their appium instance.

Amrkamel1:
> > i had to downgrade again to .17.0 to make it work
> 
> Is this Appium Desktop version?
> I thought something error returned from their appium instance.

Yes appium desk top version 1.17.1

Through error right away after u start inpection

It works on 1.17.0

Amrkamel1:
Good that you figured it out will be waiting for the fix

KazuCocoa:
@Amrkamel1 Could you try out https://github.com/appium/appium-desktop/releases/tag/v1.17.1-1 ?

Amrkamel1:
> @Amrkamel1 Could you try out https://github.com/appium/appium-desktop/releases/tag/v1.17.1-1 ?

i will and get back to you

KazuCocoa:
Let me know if the 1.17.1-1 binary also does not work

Amrkamel1:
> Let me know if the 1.17.1-1 binary also does not work
it works fine now thanks

JayasriPV:
I am using mac and installed latest version of appium desktop 1.17.1-1 still facing the issue

201905270246:
@KazuCocoa 

> Let me know if the 1.17.1-1 binary also does not work
yes, doesn't work.

mac 10.15
appium desktop 1.17.1-1

KazuCocoa:
It is appreciated debug more since I tested with my personal account before. Then, it had no issue about HTTP or HTTPS. Maybe your case was a different situation.

IsaacDempsey:
Updating to the latest version of appium desktop 1.17.1-1 fixed this issue for me also.

MyatNoeWai:
I have the same error on Appium Desktop.

The problem
- unable to start appium inspector and get "could not start session" error.
[...]

PR #1412, opened 2020-06-04
fix: npm install --save wd@1.11.2

KazuCocoa:
https://github.com/appium/appium-desktop/issues/1410

package.json has `"wd": "^1.11.2",`, but a release build was built by 1.12 which had an issue for HTTPS


KazuCocoa:
Will tag 1.17.1-1 as a hotfix after this merge

mykola-mokhnach:
Does it also make sense to lock the max wd package version in package.json?

KazuCocoa:
good idea

PR #1438, opened 2020-06-22
feat: Add Webview notification in Selected Element

wswebcreation:
In the last few months I've seen a lot of people using Appium Desktop to inspect webpages and or Webview contexts. This is not a best practice and I thought that Appium Desktop should at least warn people that it is better to switch to the webview context if they easily want to automate their browser and or hybrid app.
The ultimate goal would be to also inspect a webpage / hybrid app webview but that would be phase 2.

This PR adds:
- when Safari is started it will add `includeSafariInWebviews: true` to the capabilities so people will get the correct warning
- show a warning in the Selected element section

Here are some screenshots on how it looks like for browser, hybrid and native

Please let me know if this is a good solution or if I need to change some stuff / the message

**Safari browser**
image

**iOS Native**
image

**iOS Hybrid**
image

**Android browser**
image

**Android Native**
image

**Android Hybrid**
image

### DIFF
--- a/app/main/appium-method-handler.js
+++ b/app/main/appium-method-handler.js
@@ -109,7 +109,7 @@ export default class AppiumMethodHandler {
     return {variableName, variableType, strategy, selector, elements};
   }
 
-  async _execute ({elementId, methodName, args, skipContextsScreenshotAndSource}) {
+  async _execute ({elementId, methodName, args, skipRefresh}) {
     this._lastActiveMoment = +(new Date());
     let cachedEl;
     let res = {};
@@ -147,7 +147,7 @@ export default class AppiumMethodHandler {
     await Bluebird.delay(500);
 
     let contextsSourceAndScreenshot;
-    if (!skipContextsScreenshotAndSource) {
+    if (!skipRefresh) {
       contextsSourceAndScreenshot = await this._getContextsSourceAndScreenshot();
     }
 
@@ -158,12 +158,12 @@ export default class AppiumMethodHandler {
     };
   }
 
-  async executeElementCommand (elementId, methodName, args = [], skipContextsScreenshotAndSource = false) {
-    return await this._execute({elementId, methodName, args, skipContextsScreenshotAndSource});
+  async executeElementCommand (elementId, methodName, args = [], skipRefresh = false) {
+    return await this._execute({elementId, methodName, args, skipRefresh});
   }
 
-  async executeMethod (methodName, args = [], skipContextsScreenshotAndSource = false) {
-    return await this._execute({methodName, args, skipContextsScreenshotAndSource});
+  async executeMethod (methodName, args = [], skipRefresh = false) {
+    return await this._execute({methodName, args, skipRefresh});
   }
 
   async _getContextsSourceAndScreenshot () {
--- a/app/main/appium-method-handler.js
+++ b/app/main/appium-method-handler.js
@@ -109,7 +109,7 @@ export default class AppiumMethodHandler {
     return {variableName, variableType, strategy, selector, elements};
   }
 
-  async _execute ({elementId, methodName, args, skipScreenshotAndSource}) {
+  async _execute ({elementId, methodName, args, skipContextsScreenshotAndSource}) {
     this._lastActiveMoment = +(new Date());
     let cachedEl;
     let res = {};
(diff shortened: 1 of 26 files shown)


app/renderer/components/Inspector/Inspector.js
```diff
@@ -121,6 +121,8 @@ export default class Inspector extends Component {
            showKeepAlivePrompt, keepSessionAlive, sourceXML, t} = this.props;
     const {path} = selectedElement;
 
+
+
```
KazuCocoa, line 125: No need lines?

assets/locales/en/translation.json
```diff
@@ -170,6 +170,7 @@
   "Selector": "Selector",
   "Interactions are not available for this element": "Interactions are not available for this element",
   "usingXPathNotRecommended": "Using XPath locators is not recommended and can lead to fragile tests. Ask your development team to provide unique accessibility locators instead!",
+  "usingSwitchContextRecommended": "Multiple contexts are detected. It is advised to switch context to WEBVIEW if you easily want to select elements like you would do on the web.",
```
KazuCocoa, line 173: Only 'en' resource is fine 👍 
https://github.com/appium/appium-desktop/blob/52eb9c1c8076fe059ead31df24c62b4556e3ea73/CONTRIBUTING.md#localization

app/renderer/actions/Session.js
```diff
@@ -709,3 +710,18 @@ export function setVisibleProviders () {
...
+ * for future HTML detection
+ *
+ * @param {object} caps
+ */
+function includeSafariInWebviews (caps) {
+  const {browserName = ''} = caps;
+
+  return {
+    ...caps,
+    ...(browserName.toLowerCase() === 'safari' ? {includeSafariInWebviews: true} : {}),
```
KazuCocoa, line 725: Is it safe to check 'platformName', too?
Currently, we do not have a plan to add browser safari, but includeSafariInWebviews works only for our mobile safari.

app/main/appium-method-handler.js
```diff
@@ -109,7 +109,7 @@ export default class AppiumMethodHandler {
     return {variableName, variableType, strategy, selector, elements};
   }
 
-  async _execute ({elementId, methodName, args, skipScreenshotAndSource}) {
+  async _execute ({elementId, methodName, args, skipContextsScreenshotAndSource}) {
```
jlipps, line 112: `skipContextsScreenshotAndSource` is now getting really long, probably should refactor into a name that is short but descriptive, maybe `skipMetadataRetrieval` or just `skipRefresh`?

app/renderer/components/Inspector/Inspector.js
wswebcreation: Will remove them

app/renderer/actions/Session.js
wswebcreation: I can add that if we want to be on the safe side

wswebcreation:
@jlipps 

> I like this, but I'm wondering what the value of the warning will be if the user can't do anything about it in Appium Desktop anyway. Maybe the warning should say something like "Webview context(s) detected; certain elements might only be accessible after switching to a webview context, which must be done in an Appium script."?

I totally agree with you, but maybe that should be phase 2. My goal now was to at least warn people about it. I've seen too many customers using native selectors for web elements and getting in to trouble (slowing tests down enormously on iOS)  
I'm happy to change the message, please let me know what you all like to see

assets/locales/en/translation.json
wswebcreation: ahh, so I can remove all the other languages?

mykola-mokhnach: exactly

app/main/appium-method-handler.js
```diff
@@ -146,28 +146,38 @@ export default class AppiumMethodHandler {
...
 
-  async _getSourceAndScreenshot () {
-    let source, sourceError, screenshot, screenshotError, windowSize, windowSizeError;
+  async _getContextsSourceAndScreenshot () {
+    let contexts, contextsError, source, sourceError, screenshot, screenshotError, windowSize, windowSizeError;
+
+    try {
+      contexts = await this.driver.contexts();
+    } catch (e) {
+      if (e.status === 6) {
```
mykola-mokhnach, line 175: what is the meaning of this magic number?

app/renderer/actions/Session.js
wswebcreation: @KazuCocoa 

Was just thinking about that, but there are some cloud vendors that don't make those capabilities like platformName/version mandatory. That would break this functionality if we would add more caps like that. Can we just leave it like this?

app/main/appium-method-handler.js
wswebcreation: updated to `skipRefresh `

app/main/appium-method-handler.js
wswebcreation: @mykola-mokhnach 

Not 100% sure about the code, never seen it so just relied on the usage of if for the other commands and copied the structure

wswebcreation:
@jlipps 

I've used your text and added a link to it, see below. The downside is that the link can't be a real link, the alert component doesn't allow html links so I wasn't able to make it clickable


image

app/renderer/actions/Session.js
KazuCocoa: ah, okay. Np for current way.
I came up with safaridriver case as an example, but safaridriver should have not work with current capability since newer one only accepted W3C capabilities. So, current AD's behaviour also does not work for them directly.

jlipps:
looks great to me @wswebcreation !

KazuCocoa:
let me retry the ci

PR #1443, opened 2020-06-29
Feat:add webview inspector

wswebcreation:
This PR will add a webview inspector for Android **AND** iOS with:
- A select box for selecting the webview.
image
- Tooltip for the select box
image
- webview being shown with the title of the page, to make it easier to select the correct webview (in case of multiple). This was working out of the box for iOS but not for Android. So that has been added
image
- when you select a webview it will show a _proper_ HTML-tree, `head/script` are removed
image
- proper logic to select elements by adding `wdith/height/x/y` data on the HTML
image
- select elements when the address bar has been made smaller due to manual swiping
image
- add a warning to notify people that the Appium-Desktop webview inspector might not be that accurate
image
- it will automatically start in the _webview_-context if a session for a Chrome or Safari browser is started (removed the default which went to `NATIVE_APP`)
- it will automatically start in the _webview_-context if a session for a Hybrid app is started with `autoWebview: true`
- A pain point for a lot of users is that Appium (Desktop) breaks if they don't provide the correct ChromeDriver version. I now let Appium Desktop download it by default so AD will support Chrome and Android Hybrid apps out of the box without any errors by enabling the relaxed security flag by default. A tooltip is added with the reason why.
image
- when a browser session is started AD will automatically load a default url, which will be the AD docs. This is done because ChromeDriver starts with an empty page and that _breaks_ the logic because there is no HTML to show and users get stuck


KazuCocoa:
yey, see this pr tomorrow :)

jlipps:
don't have time to review just yet, but wow, looks amazing!

wswebcreation:
@KazuCocoa and @jlipps 

In this piece of code https://github.com/appium/appium-desktop/blob/master/app/main/appium.js#L233 we start Appium in the native context by default. Can we disable this when this PR is getting merged?

app/main/appium-method-handler.js
```diff
@@ -3,6 +3,7 @@ import wd from 'wd';
 import log from 'electron-log';
 import _ from 'lodash';
 import { SCREENSHOT_INTERACTION_MODE } from '../renderer/components/Inspector/shared';
+import {getWebviewStatusAddressBarHeight, setHtmlElementAttributes} from './helpers';
```
KazuCocoa, line 6: This import may break npm run test. I think some electron dependencies haven't been ready in the test code, but here tried to import them.

app/main/appium-method-handler.js
```diff
@@ -167,17 +168,43 @@ export default class AppiumMethodHandler {
...
+      }
+      currentContextError = e;
+    }
+
+    /**
+     * If its a webview then update the HTML with the element location
+     * so the source can be used in the native inspector
+     */
+    try {
+      if (currentContext !== 'NATIVE_APP') {
```
KazuCocoa, line 199: `NATIVE_APP` can extract as a const

app/main/appium-method-handler.js
```diff
@@ -167,17 +168,43 @@ export default class AppiumMethodHandler {
...
+        const webviewStatusAddressBarHeight = await this.driver.execute(getWebviewStatusAddressBarHeight, [{statBarHeight}]);
+
+        await this.driver.execute(setHtmlElementAttributes, [{webviewStatusAddressBarHeight}]);
+      }
+    } catch (e) {
```
KazuCocoa, line 204: We often write as `} catch (ign) {}`

app/main/appium-method-handler.js
```diff
@@ -265,4 +305,5 @@ export function getSessionHandler (winId) {
   }
 }
 
+
```
KazuCocoa, line 308: A redundant line

app/main/appium-method-handler.js
KazuCocoa: Probably a new file for getWebviewStatusAddressBarHeight and setHtmlElementAttributes and import them may fix this import error

app/renderer/actions/Inspector.js
```diff
@@ -191,14 +194,22 @@ export function applyClientMethod (params) {
       dispatch({type: METHOD_CALL_DONE});
 
       if (source && screenshot) {
+        let newSource = source;
+        if (source.includes('<html')) {
+          // Remove head and scripts which cause extra noise in the tree
+          newSource = source.replace(/<head.*?\/head>/s, '').replace(/<script\b[^<]*(?:(?!<\/script>)<[^<]*)*<\/script>/g, '');
```
KazuCocoa, line 200: Could extract this logic and left a unit test as an example?

KazuCocoa:
> In this piece of code https://github.com/appium/appium-desktop/blob/master/app/main/appium.js#L233 we start Appium-Desktop in the native context by default. Can we disable this when this PR is getting merged?

Yes, I think we can remove there.
The reason was this AD had not supported web context. By this change, the situation will change.

wswebcreation:
> > In this piece of code https://github.com/appium/appium-desktop/blob/master/app/main/appium.js#L233 we start Appium-Desktop in the native context by default. Can we disable this when this PR is getting merged?
> 
> Yes, I think we can remove there.
> The reason was this AD had not supported web context. By this change, the situation will change.

@KazuCocoa 

Yes, that did the trick, tnx

app/main/appium-method-handler.js
wswebcreation: Fixed it, thanks

app/main/appium-method-handler.js
wswebcreation: Done, thanks

app/main/appium-method-handler.js
wswebcreation: Done, thanks

app/renderer/actions/Inspector.js
wswebcreation: Removed this code and put it into `webviewHelpers.js` and used _proper_ DOM element removal

app/main/appium-method-handler.js
wswebcreation: Done, tnx

wswebcreation:
@jlipps @dpgraham @KazuCocoa 

I've finished the PR and removed the draft text. All added features are explained in the description and it works for Android and iOS.

Thanks for the review so far and please let me know if I need to do more for it.

app/main/appium.js
```diff
@@ -57,6 +57,9 @@ function connectStartServer (win) {
     // than the typical behavior, which is process.exit o_O
     args.throwInsteadOfExit = true;
 
+    // Let Appium Desktop automatically download the matching ChromeDriver
+    args.allowInsecure.push('chromedriver_autodownload');
```
KazuCocoa, line 61: Here should be configurable in the advanced tab.
Could you remove this for now or include this capability in the advanced tab?


Because some users would not download binaries automatically under a slow network, for example.
Then, this flag should be able to be disabled.
(I saw such comments)

app/main/webviewHelpers.js
```diff
@@ -0,0 +1,82 @@
+import {load} from 'cheerio';
```
KazuCocoa, line 1: 👍

app/renderer/actions/Session.js
```diff
@@ -719,9 +720,33 @@ export function setVisibleProviders () {
...
+ * all internal calls are still JSONWP calls
+ *
+ * @param {object} caps
+ */
+function setChromeDriverToJsonwp (caps) {
+  const {browserName = ''} = caps;
+  const chromeCapabilities = {
+    nativeWebScreenshot: true,
+    chromeOptions: {
+      'w3c': false,
```
KazuCocoa, line 744: ah, yeah, https://github.com/admc/wd AD uses as WebDriver client is still JSONWP

app/main/appium.js
wswebcreation: Good remark, on the other hand, this is something that would definitely break the usage for Android if the correct CrhomeDriver is not downloaded. I also wonder how many people even look into these settings.
Can we enable it by default and let people disable it if needed?

app/renderer/actions/Session.js
wswebcreation: Yep, it broke a lot for my, then I figured out this was the best way

app/main/appium.js
KazuCocoa: Yes, it seems good > turn it on by default
My main point is adding the allowInsecure in the advanced tab instead of hardcoding it (in this PR or as a separate PR)

wswebcreation: @KazuCocoa 

I just looked into it again, but AD already has `allowInsecure` enabled, that's why I pushed it (otherwise the push would fail). This is a screenshot of Appium 1.17. So is this a bug in the current version?

Last but not least, this is a 1 time thing IF they start with Chrome browser OR use a hybrid app where they switch to the webview. Should we still make it configurable (makes the code more complex with error handling and all kinds of other situations) or keep it this way. It's max 15MB that will be downloaded for a specific local device with Chrome

image

And it's disabled here 
image

KazuCocoa: I see. The default value is an empty array. When we specify arg like `--allow-insecure=foo,bar` for appium server in the launching time, it will get the body.
So, in the advanced tab, the option may be text field like server address, log level etc instead of a simple checkbox.

wswebcreation: Ahh, ok. Then I'll add a field here to automatically download ChromeDriver and enable it by default.

Do you have any other feedback on this PR that I need to fix?

wswebcreation: Fixed it with this based on @mykola-mokhnach his feedback

image

KazuCocoa: 👍 
It seems reasonable to me

app/main/appium-method-handler.js
```diff
@@ -167,46 +169,120 @@ export default class AppiumMethodHandler {
...
+    // Get the contexts and retrieve extra data
+    const contexts = await this.driver.contexts();
+    const getContextData = async (context) => {
+      let title;
+      const id = context;
+      if (id !== NATIVE_APP) {
+        await this.driver.context(context);
+        const pageTitle = await this.driver.title();
+        title = pageTitle === '' ? 'No page title available' : pageTitle;
```
KazuCocoa, line 270: `pageTitle === ''` could be `_.isEmpty(pageTitle)`  (lodash)

wswebcreation: Fixed it, thanks

PR #1414, opened 2020-06-06
feat: Add Windows 32-bit binary and zip

KazuCocoa:
Created a branch in this repo to run a package build (original https://github.com/appium/appium-desktop/pull/1251 )
Bumped the version to build a new package for testing (reverted by 66f3d354c122ab307304a60d18630eaf229bcd69)

This will increase the size of `.exe` file since it will get 32 bit binary.

PR #1448, opened 2020-07-05
feat: add locator timings

wswebcreation:
The goal of this PR is to give people insights on the time it will take to locate elements with a certain locator strategy. It will add:
- a button in the inspector window to get times of finding elements with a certain locator strategy
- a spinner for waiting
- find element times in the _Find by_ table

**Android example**
appium-times-android

**iOS example:**
appium-times-ios


app/main/appium-method-handler.js
```diff
@@ -64,7 +64,11 @@ export default class AppiumMethodHandler {
   }
 
   async fetchElement (strategy, selector) {
+    const start = process.hrtime.bigint();
```
mykola-mokhnach, line 67: Please use the Timer class from appium-support lib

mykola-mokhnach: https://github.com/appium/appium-support/blob/b90d85fde1c1ce2aee0a2511e5bbff5cc11c1549/lib/net.js#L155

app/renderer/components/Inspector/SelectedElement.js
```diff
@@ -140,30 +156,39 @@ class SelectedElement extends Component {
...
             >
               {t('Clear')}
             </Button>
+            <Tooltip title={t('getTimes')}>
+              <Button
+                disabled={isDisabled}
+                id='btnGetTimings'
+                onClick={() => getFindElementsTimes(findDataSource)}
+              >
+                {t('Get Times')}
```
mykola-mokhnach, line 186: Is this string present in resources?

mykola-mokhnach:
I think from UX perspective it is not very obvious that Get Times button should fill up these cells. Would it be possible to put such button or a link directly into the cell where the resulting number should appear?

app/main/appium-method-handler.js
wswebcreation: Thanks, will do

app/renderer/components/Inspector/SelectedElement.js
wswebcreation: No, will add it

wswebcreation:
> I think from UX perspective it is not very obvious that Get Times button should fill up these cells. Would it be possible to put such button or a link directly into the cell where the resulting number should appear?

@mykola-mokhnach 

I've added a button to make it more clear like this, the tool tip is still present.


image

image

mykola-mokhnach:
I think `Timing` would be a more precise term here. `Get Time` has a bit different meaning to me (although I'm not a native speaker, so it might be a wrong assumption)

wswebcreation:
> I think `Timing` would be a more precise term here. `Get Time` has a bit different meaning to me (although I'm not a native speaker, so it might be a wrong assumption)

Change it to this, including the tool tip

image

wswebcreation:
Tnx for merging @KazuCocoa 

Can we get a new release in the coming days with those 2 new features?

KazuCocoa:
Perhaps yes.
Once bump current version like https://github.com/appium/appium-desktop/pull/1414/commits/f0ffcddc7598f268fdf342df04b6e2ab45f5f7eb and run a CI job, the new one will override https://github.com/appium/appium-desktop/releases/tag/v1.18.0-beta.0 (Or we should increment the version)
Let me try

KazuCocoa:
@wswebcreation 
Current master is available from https://github.com/appium/appium-desktop/releases/tag/v1.18.0-beta.1 as pre release

wswebcreation:
@KazuCocoa 

Thanks!! It's working, just one question, can you also add this PR https://github.com/appium/appium-desktop/pull/1448 in the description? The `Get Timings` is also merged and in this release

KazuCocoa:
👍 added

Issue #1449, opened 2020-07-05
An unexpected error occurred with message: Cannot read property 'length' of undefined.


arunrona:
## Appium or Appium Desktop?

You are reporting an issue at the Appium Desktop repository. Appium Desktop is a wrapper around Appium. If you are having trouble running tests, it is much more likely that the problem you are encountering is not a problem with Appium Desktop but with Appium. For that reason we require you to have tried your testcase with Appium CLI before reporting issues. Check the checkbox below to confirm that you have proven your issue does _not_ reproduce on Appium itself:

- [* ] I have verified that my issue does not occur with Appium and should be investigated as an Appium Desktop issue

## The problem

_Replace this section with a description of your issue_

## Environment

- I am running Appium Desktop version _<insert version here>_.
- I am on (pick one):
    - [* ] Mac
    - [ ] Windows
    - [ ] Linux
While trying to launch my app getting below error

An unexpected error occurred with message: Cannot read property 'length' of undefined.
Please kindly open up a bug report at https://github.com/appium/appium-desktop/issues/new.
TypeError: Cannot read property 'length' of undefined
    at recursive (file:///Applications/Appium.app/Contents/Resources/app/app/renderer/components/Inspector/Source.js:71:28)
    at c.render (file:///Applications/Appium.app/Contents/Resources/app/app/renderer/components/Inspector/Source.js:90:21)
    at finishClassComponent (/Applications/Appium.app/Contents/Resources/app/node_modules/react-dom/cjs/react-dom.development.js:17160:31)
    at updateClassComponent (/Applications/Appium.app/Contents/Resources/app/node_modules/react-dom/cjs/react-dom.development.js:17110:24)
    at beginWork (/Applications/Appium.app/Contents/Resources/app/node_modules/react-dom/cjs/react-dom.development.js:18620:16)
    at HTMLUnknownElement.callCallback (/Applications/Appium.app/Contents/Resources/app/node_modules/react-dom/cjs/react-dom.development.js:188:14)
    at Object.invokeGuardedCallbackDev (/Applications/Appium.app/Contents/Resources/app/node_modules/react-dom/cjs/react-dom.development.js:237:16)
    at invokeGuardedCallback (/Applications/Appium.app/Contents/Resources/app/node_modules/react-dom/cjs/react-dom.development.js:292:31)
    at beginWork$1 (/Applications/Appium.app/Contents/Resources/app/node_modules/react-dom/cjs/react-dom.development.js:23203:7)
[...]

PR #1450, opened 2020-07-05
fix: no children in source

KazuCocoa:
fix https://github.com/appium/appium-desktop/issues/1449
`elemObj.children` could be undefined.

### DIFF
--- a/app/renderer/components/Inspector/Source.js
+++ b/app/renderer/components/Inspector/Source.js
@@ -67,9 +67,7 @@ class Source extends Component {
 
     // Recursives through the source and renders a TreeNode for an element
     let recursive = (elemObj) => {
-      if (!elemObj) {return null;}
-      if (!elemObj.children) {return null;}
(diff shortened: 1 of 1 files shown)


app/renderer/components/Inspector/Source.js
```diff
@@ -68,6 +68,7 @@ class Source extends Component {
     // Recursives through the source and renders a TreeNode for an element
     let recursive = (elemObj) => {
       if (!elemObj) {return null;}
```
mykola-mokhnach, line 70: this all could be reduced to `if (!elemObj?.children?.length) return null`

mykola-mokhnach:
It looks like optional chaining is not enabled for AD :(
Would it be possible to enable it somewhere in transpilator settings?

Otherwise we could still use `if (!((elemObj || {}).children || []).length) return null`

KazuCocoa:
ah~, gotcha. On my local, it looked transpiled 
Let me take a look as another PR :)

PR #1452, opened 2020-07-06
Feat/add new locators

wswebcreation:
XPATH is very slow for especially iOS native apps. This PR will add the following:

- add `-ios class chain`-locator based on the current XML
- add `-ios predicate string`-locator based on the current XML
- sort the table based on the _fastest_ find element locator strategy if users click on `Get Timings`
- add a tooltip and a link to the `-ios class chain` documentation

Here's an example of how it works
ad-add-selectors


mykola-mokhnach:
What is shown there if the target platform is not iOS?

app/renderer/components/Inspector/shared.js
```diff
@@ -47,6 +47,55 @@ export function getLocators (attributes, sourceXML) {
   return res;
 }
 
+export function xpathToClassChain (xpath) {
```
mykola-mokhnach, line 50: It makes sense to cover this method with unit tests

app/renderer/components/Inspector/shared.js
```diff
@@ -47,6 +47,55 @@ export function getLocators (attributes, sourceXML) {
...
+  let xpathArray = xpath
+    // XPATH can hold parentheses like this `(//XCUIElementTypeOther[@name="something"])[2]/(XCUIElementTypeOther)[4]/XCUIElementTypeOther[2]`
```
mykola-mokhnach, line 52: the transformation task is much more complicated that it could look like. The most appropriate way of doing that would be to convert an xpath expression to its AST representation and then build class chain based on this tree.

Or build a class chain based on the current element position in the source tree, similar to what we do for xpath

wswebcreation:
@mykola-mokhnach 

> What is shown there if the target platform is not iOS?

Nothing, it is only added when this condition is true `if (xpath.includes('XCUIElement') && currentContext === NATIVE_APP)`

app/renderer/components/Inspector/shared.js
wswebcreation: Totally agree, didn't want to do that yet to see how you all feel about it

app/renderer/components/Inspector/shared.js
wswebcreation: @mykola-mokhnach 

You mean something like this https://github.com/appium/appium-desktop/blob/master/app/renderer/util.js#L68 ?

I only wonder how we should do that if we only have the xml tree (but I could be wrong)

KazuCocoa:
As another thing, we can allow users to modify the class chain and can run it with 'get timing' for a case a user wants to optimise the locator path. (not for this PR)

wswebcreation:
@KazuCocoa 

> As another thing, we can allow users to modify the class chain and can run it with 'get timing' for a case a user wants to optimise the locator path. (not for this PR)

We already have the search modal for that, maybe we can also add timing there as a default?

KazuCocoa:
it's a good idea

wswebcreation:
@KazuCocoa and @mykola-mokhnach 

I've refactored the way the `-ios class chain` is being created and I've added the `-ios predicate string` to it, see below. I already had a conversation with @mykola-mokhnach to refactor the code better, because there is a lot of duplication. On the other hand, the code is so complex that I'm worried that I might break more then I would like.

ad-add-selectors

wswebcreation:
@KazuCocoa and @mykola-mokhnach 

Just checking, but do you have any feedback on this PR

app/renderer/util.js
```diff
@@ -121,6 +137,119 @@ export function getOptimalXPath (doc, domNode, uniqueAttributes = ['id']) {
...
+    // which is not an official XCUITest element, return empty string
+    if (!domNode.tagName || domNode.nodeType !== 1 || domNode.tagName === 'XCUIElementTypeApplication') {
+      return '';
+    }
+
+    // BASE CASE #2: If this node has a unique class chain based on attributes then return it
+    for (let attrName of uniqueAttributes) {
+      const attrValue = domNode.getAttribute(attrName);
+      if (attrValue) {
+        let xpath = `//${domNode.tagName || '*'}[@${attrName}="${attrValue}"]`;
```
mykola-mokhnach, line 160: I personally prefer using `const` everywhere for variables whose values are not going to be changed

app/renderer/util.js
```diff
@@ -121,6 +137,119 @@ export function getOptimalXPath (doc, domNode, uniqueAttributes = ['id']) {
...
+        let index = childNodes.indexOf(domNode);
+        classChain += `[${index + 1}]`;
+      }
+    }
+
+    // Make a recursive call to this nodes parents and prepend it to this xpath
+    return getOptimalClassChain(doc, domNode.parentNode, uniqueAttributes) + classChain;
+  } catch (ign) {
+    // If there's an unexpected exception, abort and don't get an XPath
+    return null;
```
mykola-mokhnach, line 201: I assume it makes sense to write the exception into the debug log. Otherwise it would be very hard to catch it

app/renderer/util.js
```diff
@@ -121,6 +137,119 @@ export function getOptimalXPath (doc, domNode, uniqueAttributes = ['id']) {
...
+      return '';
+    }
+
+    // BASE CASE #2: Check all attributes and try to find the best way
+    let xpathAttributes = [];
+    let predicateString = [];
+
+    for (let attrName of uniqueAttributes) {
+      const attrValue = domNode.getAttribute(attrName);
+      if (attrValue) {
```
mykola-mokhnach, line 228: `if _,isNil(atttrValue) continue`

app/renderer/util.js
```diff
@@ -121,6 +137,119 @@ export function getOptimalXPath (doc, domNode, uniqueAttributes = ['id']) {
...
+        // If the XPath does not parse, move to the next unique attribute
+        try {
+          othersWithAttr = XPath.select(xpath, doc);
+        } catch (ign) {
+          continue;
+        }
+
+        // If the attribute isn't actually unique, get it's index too
+        if (othersWithAttr.length === 1) {
+          return predicateString.join(' AND ');
```
mykola-mokhnach, line 243: There might be the case where all attributes match, so it is not possible to create a locator, that uniquely identifies the element

app/renderer/util.js
```diff
@@ -121,6 +137,119 @@ export function getOptimalXPath (doc, domNode, uniqueAttributes = ['id']) {
...
+        }
+      }
+    }
+  } catch (ign) {
+    // If there's an unexpected exception, abort and don't get an XPath
+    return null;
```
mykola-mokhnach, line 249: ⬆️

test/unit/util-specs.js
```diff
@@ -49,6 +49,8 @@ describe('util.js', function () {
           },
           xpath: '//XCUIElementTypeApplication[@name="🦋"]/XCUIElementTypeWindow',
           path: '0',
+          classChain: '**/XCUIElementTypeWindow',
```
mykola-mokhnach, line 52: It would be good to have more tests. At least one where indexing is present and the second where more that one attribute is included

app/renderer/util.js
wswebcreation: Fixed it, thanks

app/renderer/util.js
wswebcreation: Fixed it, thanks

app/renderer/util.js
wswebcreation: Not 100% sure what you mean with this, can you explain?

app/renderer/util.js
wswebcreation: Think I fixed this

test/unit/util-specs.js
wswebcreation: Will add a UT for that

app/renderer/util.js
wswebcreation: @mykola-mokhnach 

Can you check if this is what you mean?

test/unit/util-specs.js
wswebcreation: @mykola-mokhnach 
Added 2 tests, does it fit what you were expecting

app/renderer/util.js
mykola-mokhnach: I mean there might be elements whose all attributes match, but they are not the same. Which means the resulting locator might not be selecting what is highlighted in the inspector

app/renderer/components/Inspector/SelectedElement.js
```diff
@@ -141,6 +143,48 @@ class SelectedElement extends Component {
...
+        key: '-ios class chain',
+        find: classChainText,
+        selector: classChain,
+        time: getTimeButton
+      });
+    }
+
+    // Add predicate string to the data source as well
+    if (predicateString && currentContext === NATIVE_APP) {
+      const predicateStringText = <Tooltip title={t('This selector is in BETA, it is the XML selector translated to `-ios predicate string`.')}>
```
mykola-mokhnach, line 169: is this tooltip still relevant?

app/renderer/components/Inspector/shared.js
mykola-mokhnach: I mean something more advanced, but the current implementation is fine as for POC

app/renderer/util.js
mykola-mokhnach: I don't observe anything is done. I was expected to see something like `log.debug(e.stack)`

app/renderer/components/Inspector/SelectedElement.js
wswebcreation: I think so, it's still a beta, but happy to remove it or change the description if needed. Same is for the `class chain` selector, that one has the same type of description

app/renderer/util.js
wswebcreation: Hmm I changed it to returning an empty string/. But the `return null` is exactly the same implementation as the original `getOptimalXPath` which I didn't change

app/renderer/util.js
wswebcreation: That's true, but to be honest, I'm not 100% sure on how to overcome that. Do you have an idea?

app/renderer/util.js
wswebcreation: This is also copied form the `getOptimalXPath` methods which also returned `null`, to keep it aligned

mykola-mokhnach: the returned value is fine. I am concerned about logging the exception stack trace rather than silencing it completely

app/renderer/util.js
mykola-mokhnach: we could run the xpath equivalent to see if it matches more than one element. Although, I'm not sure about what to show in the table cell in such case

wswebcreation: @mykola-mokhnach 

That is already checked at line 236 and then 242. Meaning the predicate string is translated to XPATH to validate it. Is that sufficient?

app/renderer/util.js
wswebcreation: @mykola-mokhnach 

Added error log now, strange this is that you can't see it here, but you can see it in my latest commit.

app/renderer/util.js
```diff
@@ -115,8 +132,130 @@ export function getOptimalXPath (doc, domNode, uniqueAttributes = ['id']) {
 
     // Make a recursive call to this nodes parents and prepend it to this xpath
     return getOptimalXPath(doc, domNode.parentNode, uniqueAttributes) + xpath;
-  } catch (ign) {
+  } catch (error) {
+    // If there's an unexpected exception, abort and don't get an XPath
+    log.error(`The most optimal XPATH could not be determined because an error was thrown: '${JSON.stringify(error, null, 2)}'`);
```
mykola-mokhnach, line 137: `${JSON.stringify(error, null, 2)}` -> error.stack

app/renderer/util.js
mykola-mokhnach: ah, true. In such case the loop would never return and finally undefined would be returned. Makes sense to me

wswebcreation:
Thanks for all the support and feedback on this PR @mykola-mokhnach and @KazuCocoa

PR #1462, opened 2020-07-09
fix: keep compatibility in _getContextsSourceAndScreenshot for platforms which have no context commands

KazuCocoa:
https://github.com/appium/appium-desktop/pull/1438 introduced context-related commands, but some platform does not support them like Windows and macOS (https://github.com/appium/appium-desktop/issues/1449#issuecomment-655966896) .
We should keep the previous way for such platforms.

In this PR, AD will follow the previous logic if the app under test does not have get contexts command.
I also fixed some errors I found in tests.

### DIFF
--- a/app/main/appium-method-handler.js
+++ b/app/main/appium-method-handler.js
@@ -184,7 +184,7 @@ export default class AppiumMethodHandler {
    * @returns {boolean} True if the platformName support context switch
    *
    */
-  _hasContextSwitch (platformName) {
+  _hasContextCommands (platformName) {
     return (['android', 'ios'].includes(_.toLower(platformName)));
   }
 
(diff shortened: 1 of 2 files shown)


KazuCocoa:
cc @wswebcreation

app/main/appium-method-handler.js
```diff
@@ -174,60 +174,84 @@ export default class AppiumMethodHandler {
...
         source, sourceError, screenshot, screenshotError, statBarHeight, windowSize, windowSizeError;
 
+    // Get current session capabilities is not in W3C spec,
+    // so potentially the request fails. Then, let's ignore the result.
     try {
-      currentContext = await this.driver.currentContext();
-    } catch (e) {
-      if (e.status === 6) {
-        throw e;
+      ({platformName, statBarHeight} = await this.driver.sessionCapabilities());
```
wswebcreation, line 198: @KazuCocoa 

You placed the command here now, it was lower and executed in the native context and will _break_ for Android. Let me check this for sure

wswebcreation: @KazuCocoa 

Yes, this fix breaks element inspection on Chrome browser. The position of the elements can't be determined so the can't be added in the screenshot.

**Before the fix:**
image

**After the fix:**
image

The order in which the commands are now executed are not correct

KazuCocoa:
thanks for checking :)

Issue #1470, opened 2020-07-17
Inspector does not provide error message


DmytroPazykovRing:
## The problem

After submitting an incorrect selector to the inspector,  error or stack trace is missing. All we have is an error that return value isn't an Array, which is expected, as soon as it is probably a String error or something.

## Way to reproduce

- Attach to a real session
- Open Search for Element 
- Choose a selector strategy - Name, UIAutomater, DataMatcher
- Input to the selector random text - asd
- Click the Search button.

_The error that will be displayed(for DataMatcher as an example):_
```
Call to '10' failed
[elements("-android datamatcher","asd")] Response value field is not an Array.
```

## Environment

- Appium Desctop version - 1.17.1-1 (Electron: 7.2.4, Node.js: 12.8.1)
- I am on (pick one):
    - [+] Mac
    - [ ] Windows
    - [ ] Linux

PR #1474, opened 2020-07-19
fix: print the original error response if given error message has the original error for findElements

KazuCocoa:
AD currently shows below message as the result of findElements via an icon when the given parameter is an invalid selector.
For example, my below screenshot is for macOS, but I sent a selector for Android.

```
Call to '10' failed
[elements("-android datamatcher","asd")] Response value field is not an Array.
```

I could not find `findElement` request via icon and actions. So, I only tweaked this case since I could not replicate an error case for `findElement`.

image

This change does not affect a case which it returns `[]`.

---

closes https://github.com/appium/appium-desktop/issues/1470


DmytroPazykovRing:
Thanks!

PR #1490, opened 2020-07-23
ci: introduce GitHub Actions for weekly translation job

KazuCocoa:
This PR replaces current weekly translation task from Azure to GitHub Actions.

For now, the azure task failed because of github token issue but we no longer take care about it with this change.

https://github.com/appium/appium-desktop/settings/secrets has `CROWDIN_PROJECT_KEY`
Below is a screenshot the sync worked.
image

https://github.com/appium/appium-desktop/pull/1488 is an example to create a PR by this job. If no diff exists, it won't create a new PR.

----

https://github.com/appium/appium-desktop/runs/902100888?check_suite_focus=true is the latest example job


.github/workflows/translation.yml
```diff
@@ -1,4 +1,4 @@
-name: Weekly job
+name: Weekly Sync Translations job
 
 on:
   workflow_dispatch:
```
mykola-mokhnach, line 4: I think it makes sense to replace the last 0 with random minute (usually `H`)

Issue #1505, opened 2020-08-04
Can't attach to the session, the button looks active but unresponsive


n3gbx:
## Appium or Appium Desktop?

- [x] I have verified that my issue does not occur with Appium and should be investigated as an Appium Desktop issue

## The problem

Can't attach to existent session by clicking on corresponding button: the session is detected successfully, but clicking the **'Attach to Session'** button doesn't do anything.

## Environment

- I am running Appium Desktop version 1.18.0.
- I am on:
    - [ ] Mac
    - [ ] Windows
    - [x] Linux (Ubuntu 18.04.4)

priyankshah217:
@n3gbx I am also facing the same issue, but on mac.

alex85vol:
Yeap, the same issue on mac

bhanteem:
I came on here to post the same issue. I ended up going back to 1.17.1 because of this.You can see the server isn't detecting the "Attach to session" button, because nothing happens on the server side when you click the button.

KazuCocoa:
You can try out https://github.com/appium/appium-desktop/releases/tag/v1.18.0-1

samarendras:
I am still facing issue looks like appium is not responding means it's not doing any action.

samarendras:
Simple it's navigating to the url but not doing any action in chrome when it's connected wifi.

KazuCocoa:
Are you talking about 'attack to an existing session' stuff?
If no, please create a new issue with logs.

samarendras:
I have created new issues with logs.Please help me out @KazuCocoa .

NSDinesh:
Hi,
Still I am facing this issue
Approach 1 followed:
1.) Ran script in TestObject(Real device automation)
2.) Selected TestObject tab from Appium Desktop
3.) Provided the TestObject API key and selected the DataCentre
4.) Selected the Attach to Session
5.) Selected the dropdown - No session details populated
6.) Tried adding the session id manually. 'Attach to Session' button remains in disabled state 
Note: On selecting refresh or clicking enter after providing session id - In both case session id getting removed and  'Attach to Session' button remains in disabled state 

![image](https://user-images.githubusercontent.com/34602860/90029517-93dcf700-dcd8-11ea-85ce-a0df7c905975.png)

KazuCocoa:
Your issue is probably https://github.com/appium/appium-desktop/issues/1344
This issue was not typing by hand

NSDinesh:
@KazuCocoa 
[...]

PR #1506, opened 2020-08-04
fix: attach to an existing session

KazuCocoa:
Closes https://github.com/appium/appium-desktop/issues/1505

Attach to an existing session is `newSession(null, attachSessId)`. Then, `addCustomCaps(null)` raises an error in `const {browserName = '', platformName = ''} = caps;`. `{}` works both cases.


mykola-mokhnach:
It would be interesting to know why semantic PR plugins complains

KazuCocoa:
It was probably because this PR had only a commit, and the naming was `fix` instead of `fix:`...
Probably once I commit a new change to this PR, the PR plugin will refer to this PR's title

Issue #1492, opened 2020-07-24
PATH selection for APP while setting desired capabilities its not working from 1.16.0/1.17.0/1.18.0 beat1


DarshanKamble:
## Appium or Appium Desktop?

You are reporting an issue at the Appium Desktop repository. Appium Desktop is a wrapper around Appium. If you are having trouble running tests, it is much more likely that the problem you are encountering is not a problem with Appium Desktop but with Appium. For that reason we require you to have tried your testcase with Appium CLI before reporting issues. Check the checkbox below to confirm that you have proven your issue does _not_ reproduce on Appium itself:

- [ ] I have verified that my issue does not occur with Appium and should be investigated as an Appium Desktop issue

## The problem

PATH selection for APP while setting desired capabilities its not working from 1.16.0/1.17.0/1.18.0 beat1

## Environment

- I am running Appium Desktop version _<insert version here>_.
- I am on (pick one):
    - [ ] Mac
    - [x] Windows
    - [ ] Linux

KazuCocoa:
Do you mean `app` capability as a local path?
Could you share the full server log?

yuowo:
In Appium Desktop, when the Desired Capabilities add type is filepath, click value and select the file and open it, but it does not seem to take effect
macOS：10.15.6

```
[Appium] Welcome to Appium v1.18.0
[Appium] Non-default server args:
[Appium]   relaxedSecurityEnabled: true
[Appium]   allowInsecure: {
[Appium]   }
[Appium]   denyInsecure: {
[Appium]   }
[Appium] Appium REST http interface listener started on 0.0.0.0:4723
[HTTP] --> GET /wd/hub/sessions
[HTTP] {}
[GENERIC] Calling AppiumDriver.getSessions() with args: []
[GENERIC] Responding to client with driver.getSessions() result: []
[HTTP] <-- GET /wd/hub/sessions 200 4 ms - 40
[HTTP] 
```

KazuCocoa:
I see. probably `filpath` item does not work

DarshanKamble:
YES, its observed on both window system and Mac also
For More Reference  screenshots attached


*Thanks,*
*Darshan Kamble*
Software Developer
Balaji Software
+91-22-42642018
darshan.kamble@balajisoftware.com

PR #1521, opened 2020-08-15
fix: get filePaths

KazuCocoa:
closes https://github.com/appium/appium-desktop/issues/1492

Not sure since when, but `showOpenDialog` should be `then` syntax.
`this.handleSetType` is always `null`, so cannot bind.

After:
image

https://github.com/electron/electron/blob/v6.0.0/docs/api/dialog.md#dialogshowopendialogbrowserwindow-options

### DIFF
--- a/app/renderer/components/Session/CapabilityControl.js
+++ b/app/renderer/components/Session/CapabilityControl.js
@@ -4,27 +4,28 @@ import SessionStyles from './Session.css';
 import { remote } from 'electron';
 import { FileOutlined } from '@ant-design/icons';
 import { INPUT } from '../AntdTypes';
+import _ from 'lodash';
+import log from 'electron-log';
 
 const {dialog} = remote;
 
 export default class NewSessionForm extends Component {
 
-  getLocalFilePath (success) {
-    dialog.showOpenDialog({
-      properties: ['openFile'],
-    }).then((result) => {
-      if (result.filePaths) {
-        success(result.filePaths);
-      }
-    }).catch(/* ignore */);
-  }
-
   render () {
     const {cap, onSetCapabilityParam, isEditingDesiredCaps, id, t} = this.props;
 
-    const buttonAfter = <FileOutlined
+    const buttonAfter = (currentPath) => <FileOutlined
       className={SessionStyles['filepath-button']}
-      onClick={() => this.getLocalFilePath((filePaths) => onSetCapabilityParam(filePaths[0]))} />;
+      onClick={async () => {
+        try {
+          const {canceled, filePaths} = await dialog.showOpenDialog({properties: ['openFile']});
+          let returnPath = currentPath;
+          if (!canceled && !_.isEmpty(filePaths)) {returnPath = filePaths[0];}
+          onSetCapabilityParam(returnPath);
+        } catch (e) {
+          log.error(e);
+        }
+      }} />;
 
     switch (cap.type) {
       case 'text': return <Input disabled={isEditingDesiredCaps} id={id} placeholder={t('Value')} value={cap.value} onChange={(e) => onSetCapabilityParam(e.target.value)} />;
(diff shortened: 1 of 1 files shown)


app/renderer/components/Session/CapabilityControl.js
```diff
@@ -7,24 +7,24 @@ import { INPUT } from '../AntdTypes';
...
+  async getLocalFilePath (success) {
+    await dialog.showOpenDialog({
+      properties: ['openFile'],
+    }).then((files) => {
+      if (files.filePaths) {
+        success(files.filePaths);
       }
-    });
-    this.handleSetType = this.handleSetType.bind(this);
+    }).catch(/* ignore */);
```
mykola-mokhnach, line 19: I assume catch must expect an empty lambda as an argument (`() => {}`)

app/renderer/components/Session/CapabilityControl.js
```diff
@@ -7,24 +7,24 @@ import { INPUT } from '../AntdTypes';
...
 export default class NewSessionForm extends Component {
 
-  getLocalFilePath (success) {
-    dialog.showOpenDialog((filepath) => {
-      if (filepath) {
-        success(filepath);
+  async getLocalFilePath (success) {
+    await dialog.showOpenDialog({
+      properties: ['openFile'],
+    }).then((files) => {
```
mykola-mokhnach, line 15: I'm not quite sure how it works, but according to [this](https://www.electronjs.org/docs/api/dialog) tutorial the returned value is an object with canceled, filePaths and bookmarks properties

app/renderer/components/Session/CapabilityControl.js
mykola-mokhnach: also it is a good practice to log errors instead of ignoring them completely

app/renderer/components/Session/CapabilityControl.js
```diff
@@ -7,24 +7,24 @@ import { INPUT } from '../AntdTypes';
...
+      if (files.filePaths) {
+        success(files.filePaths);
       }
-    });
-    this.handleSetType = this.handleSetType.bind(this);
```
mykola-mokhnach, line 19: don't we need this handler anymore?

app/renderer/components/Session/CapabilityControl.js
mykola-mokhnach: I would try to rewrite this method as:

```
let canceled;
let filePaths;
try {
  ({canceled, filePaths} = await dialog.showOpenDialog(mainWindow, {
    properties: ['openFile']
  }));
} catch (e) {
  log.error(e);
}
if (!canceled && !_.isEmpty(filePaths)) {
  success(filePaths);
}
```

app/renderer/components/Session/CapabilityControl.js
```diff
@@ -7,24 +7,24 @@ import { INPUT } from '../AntdTypes';
...
+    }).catch(/* ignore */);
   }
 
   render () {
     const {cap, onSetCapabilityParam, isEditingDesiredCaps, id, t} = this.props;
 
     const buttonAfter = <FileOutlined
       className={SessionStyles['filepath-button']}
-      onClick={() => this.getLocalFilePath((filepath) => onSetCapabilityParam(filepath[0]))} />;
+      onClick={() => this.getLocalFilePath((filePaths) => onSetCapabilityParam(filePaths[0]))} />;
```
mykola-mokhnach, line 27: nit: I personally find it hard to read when functions like these are being passed as arguments. It would be easier to parse this piece of the code if we had something like

```
onClick={async () => {
  const filePaths = await this.getLocalFilePath();
  if (filePaths) {
    onSetCapabilityParam(filePaths[0]);
  }
} 
```

app/renderer/components/Session/CapabilityControl.js
KazuCocoa: Seems no, we probably do not need it. `this.handleSetType` was `nil` here so `.bind` raised an error

app/renderer/components/Session/CapabilityControl.js
```diff
@@ -4,27 +4,33 @@ import SessionStyles from './Session.css';
...
   render () {
     const {cap, onSetCapabilityParam, isEditingDesiredCaps, id, t} = this.props;
 
-    const buttonAfter = <FileOutlined
+    const buttonAfter = (currentPath) => <FileOutlined
       className={SessionStyles['filepath-button']}
-      onClick={() => this.getLocalFilePath((filepath) => onSetCapabilityParam(filepath[0]))} />;
+      onClick={async () => {
+        const filePath = await this.getLocalFilePath();
+        onSetCapabilityParam(filePath ? filePath : currentPath) ;
```
mykola-mokhnach, line 32: could be simplified to `filePath || currentPath`

mykola-mokhnach: what is the purpose of calling onSetCapabilityParam with `currentPath` argument?

KazuCocoa: To keep the current inputted path when file browsing was cancelled. Previous filepath behaviour (before #1492) also behaved as the same.

PR #1532, opened 2020-08-29
feat: add Sauce Labs Unified Platform endpoint for US

wswebcreation:
This PR will:

- add the new Sauce Labs Unified Platform US endpoint as the default
- fixes the layout to a normal row
image
- Adds a tooltip with information on how to use the old endpoint
image


sceiler:
I would change the first sentence in the tooltip to “...Unified Platform US endpoint by default”.

KazuCocoa (approved):

PR #1584, opened 2020-10-06
feat: move to web-based appium client

jlipps:
with lots of other refactoring.

plus a new feature: only look for contexts when the user opts into 'web/hybrid' mode with a new toggle (it's too slow for the general case sometimes).


app/main/appium.js
```diff
@@ -150,9 +147,8 @@ export function createNewSessionWindow (win) {
   });
 
   // When you close the session window, kill its associated Appium session (if there is one)
-  let sessionID = sessionWin.webContents.id;
   sessionWin.on('closed', () => {
-    killSession(sessionID);
+    // TODO W2D use IPC to trigger a delete session in renderer
```
mykola-mokhnach, line 151: should the TODO stay?

app/renderer/actions/Inspector.js
```diff
@@ -106,7 +89,7 @@ const findElement = _.debounce(async function (strategyMap, dispatch, getState,
     let {elementId, variableName, variableType} = await callClientMethod({
       strategy,
       selector,
-    });
+    })(dispatch, getState);
```
mykola-mokhnach, line 92: Would it be possible to make this call more explicit? To me the `await f()()` does not seem very readable

app/renderer/actions/Session.js
```diff
@@ -588,19 +610,16 @@ export function getRunningSessions () {
...
+      return;
+    }
+
+    const {hostname, port, path, ssl} = serverInfo;
+    try {
+      const adjPath = path.endsWith('/') ? path : `${path}/`;
+      const res = await ky(`http${ssl ? 's' : ''}://${hostname}:${port}${adjPath}sessions`).json();
+      dispatch({type: GET_SESSIONS_DONE, sessions: res.value});
+    } catch (ign) {
+      dispatch({type: GET_SESSIONS_DONE});
```
mykola-mokhnach, line 622: should we completely silence the error here?

app/renderer/lib/appium-client.js
```diff
@@ -0,0 +1,341 @@
...
+    return newContexts;
+  }
+
+  /**
+   * Retrieve available contexts, along with the title associated with each webview
+   */
+  async getContexts (platform) {
+    return _.toLower(platform) === 'ios'
+      ? await this.driver.executeScript('mobile:getContexts', [])
+      : await this.getAndroidContexts();
```
mykola-mokhnach, line 332: `mobile:getContexts` has also been recently added to Android

app/renderer/actions/Inspector.js
```diff
@@ -500,3 +505,98 @@ export function setActionArg (index, value) {
     dispatch({type: SET_ACTION_ARG, index, value});
   };
 }
+
+/**
+ * Ping server every 30 seconds to prevent `newCommandTimeout` from killing session
```
mykola-mokhnach, line 510: We set `newCommandTimeout` to zero by default, so I'm wondering why this method is still needed. It would be logically to assume that if a user still sets `newCommandTimeout` then he expects a session to be terminated in case of inactivity

app/main/appium.js
jlipps: Ah I missed that, will add the functionality before merging

app/renderer/actions/Inspector.js
jlipps: This is unfortunately the pattern we use everywhere anytime we call an action from another action, since we use redux's thunk model

app/renderer/actions/Session.js
jlipps: I guess we could log it

app/renderer/lib/appium-client.js
jlipps: Oh great!

app/renderer/actions/Inspector.js
jlipps: I wondered about this too, my guess was that it's for cloud services that don't support the cap. I'd love to get rid of this logic though!

app/renderer/actions/Inspector.js
mykola-mokhnach: perhaps we could just do something like

```
const action = callClientMethod({
      strategy,
      selector,
    });
await action(dispatch, getState);
```

to improve the readability?

app/main/appium.js
jlipps: ok, added session deletion on window close in the renderer

app/renderer/actions/Inspector.js
jlipps: @dpgraham can you remind us why we have this keepalive logic?

app/renderer/actions/Inspector.js
```diff
@@ -172,19 +152,21 @@ export function unselectHoveredElement (path) {
...
              screenshotError, windowSizeError, variableName,
-             variableIndex, strategy, selector} = await callClientMethod(params);
+             variableIndex, strategy, selector} = await callAction(dispatch, getState);
 
       if (isRecording) {
         // Add 'findAndAssign' line of code. Don't do it for arrays though. Arrays already have 'find' expression
         if (strategy && selector && !variableIndex && variableIndex !== 0) {
-          findAndAssign(strategy, selector, variableName, false)(dispatch, getState);
+          const findAction = findAndAssign(strategy, selector, variableName, false);
+          findAction(dispatch, getState);
```
mykola-mokhnach, line 169: should it be awaited?

app/renderer/actions/Inspector.js
```diff
@@ -255,9 +238,21 @@ export function setExpandedPaths (paths) {
 /**
  * Quit the session and go back to the new session window
  */
-export function quitSession () {
-  return async (dispatch) => {
-    await applyClientMethod({methodName: 'quit'})(dispatch);
+export function quitSession (reason, killedByUser = true) {
+  return async (dispatch, getState) => {
+    const killAction = killKeepAliveLoop();
+    killAction(dispatch, getState);
```
mykola-mokhnach, line 244: should it be awaited?

app/renderer/actions/Session.js
```diff
@@ -350,54 +350,77 @@ export function newSession (caps, attachSessId = null) {
...
+      // Save the current server settings
+      await settings.set(SESSION_SERVER_PARAMS, session.server);
+    }
+
+    // The homepage arg in ChromeDriver is not working with Appium. iOS can have a default url, but
+    // we want to keep the process equal to prevent complexity so we launch a default url here to make
+    // sure we don't start with an empty page which will not show proper HTML in the inspector
+    const {browserName = ''} = desiredCapabilities;
+
+    if (browserName.toLowerCase() !== '') {
```
mykola-mokhnach, line 405: toLowerCase is redundant here. Perhaps, `trim` would make more sense

app/renderer/actions/Session.js
```diff
@@ -350,54 +350,77 @@ export function newSession (caps, attachSessId = null) {
...
-        accessKey,
-        https,
-      })(dispatch);
-      removeNewSessionListeners();
-      dispatch(push('/inspector'));
-    });
-
-    // Save the current server settings
-    await settings.set(SESSION_SERVER_PARAMS, session.server);
+    action(dispatch);
```
mykola-mokhnach, line 422: await ?

app/renderer/actions/Session.js
```diff
@@ -514,7 +538,8 @@ export function changeServerType (serverType) {
   return async (dispatch, getState) => {
     await settings.set(SESSION_SERVER_TYPE, serverType);
     dispatch({type: CHANGE_SERVER_TYPE, serverType});
-    getRunningSessions()(dispatch, getState);
+    const action = getRunningSessions();
+    action(dispatch, getState);
```
mykola-mokhnach, line 542: await?

app/renderer/actions/Session.js
```diff
@@ -751,3 +766,18 @@ function addCustomCaps (caps) {
...
+
+export function bindWindowClose () {
+  return (dispatch, getState) => {
+    window.addEventListener('beforeunload', async (evt) => {
+      const {driver} = getState().inspector;
+      if (driver) {
+        try {
+          const action = quitSession('Window closed');
+          await action(dispatch, getState);
+        } catch (ign) {}
```
mykola-mokhnach, line 778: should we also set driver to null?

app/renderer/actions/Inspector.js
```diff
@@ -500,3 +517,103 @@ export function setActionArg (index, value) {
...
+    dispatch({type: SET_LAST_ACTIVE_MOMENT, lastActiveMoment: +(new Date())});
+    const {driver} = getState().inspector;
+
+    const keepAliveInterval = setInterval(async () => {
+      const {lastActiveMoment} = getState().inspector;
+      console.log('Pinging Appium server to keep session active'); // eslint-disable-line no-console
+      try {
+        await driver.getTimeouts(); // Pings the Appium server to keep it alive
+      } catch (ign) {}
+      const now = +(new Date());
```
mykola-mokhnach, line 535: why not to use Date.now like below?

app/renderer/components/Inspector/shared.js
```diff
@@ -135,12 +140,10 @@ export const actionDefinitions = {
   },
   'Session': {
     'Session Capabilities': {
-      'Get Session Capabilities': {methodName: 'sessionCapabilities'}
+      'Get Session Capabilities': {methodName: 'getSession'}
```
KazuCocoa, line 143: Does this work on web2driver? (just a question since this is not in w3c. I wondered if the client had this)

app/renderer/actions/Inspector.js
dpgraham: @mykola-mokhnach redux actions don't need to be awaited, you can just fire-and-forget them.

jlipps: yes, i didn't change 'await' status of any of these calls. many of them were intentionally not awaited, they just exist to trigger actions, and there is a sort of event/callback system that happens from there

app/renderer/actions/Session.js
jlipps: yeah this code was here before, but i'll fix it.

app/renderer/actions/Session.js
jlipps: it doesn't matter; when the window is closed all these JS objects will be destroyed

app/renderer/actions/Inspector.js
jlipps: because I was just moving the existing code! but i can update to keep it consistent.

app/renderer/components/Inspector/shared.js
jlipps: with appium, web2driver includes all the old protocols as well, but the w3c versions are the ones that are used if there is a clash.

app/renderer/actions/Inspector.js
dpgraham: @jlipps yeah that's because certain cloud providers shut down, regardless of if they set `newCommandTimeout`.

dpgraham (approved):

PR #1691, opened 2021-02-11
fix: getRunningSessions for BrowserStack & other cloud providers that use basicAuth

huzaifaiftikhar:
1. The "Attach to session" tab of the inspector window doesn't list the running sessions for BrowserStack, this happens because the BrowserStack APIs require the user credentials otherwise it returns a response code of 401.
2. This PR adds the `Authorization` header for the `/wd/hub/sessions` requests.

PR #1693, opened 2021-02-12
fix: getCurrentSessions for cloud providers using Basic Auth

huzaifaiftikhar:
1. The "Attach to session" tab of the inspector app doesn't return the running sessions for BrowserStack, this happens because the BrowserStack APIs require the user credentials otherwise it returns a response code of 401.
2. This PR adds the Authorization header for the `/wd/hub/sessions` requests i.e. for `getCurrentSessions`
3. Also add the ability to use the `BROWSERSTACK_PORT` set in the environment variables for defining the port and setting SSL to true if the port is 443.

### DIFF
--- a/app/main/appium.js
+++ b/app/main/appium.js
@@ -330,9 +330,15 @@ function connectClientMethodListener () {
 }
 
 const getCurrentSessions = _.debounce(async (evt, data) => {
-  const {host, port, path: appiumPath = '/wd/hub', ssl} = data;
+  const {host, port, path: appiumPath = '/wd/hub', ssl, username, accessKey} = data;
+  let res;
   try {
-    const res = await request(`http${ssl ? 's' : ''}://${host}:${port}${appiumPath}/sessions`);
+    if (username && accessKey) {
+      // Basic Authorization for some cloud providers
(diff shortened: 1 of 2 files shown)


huzaifaiftikhar:
@KazuCocoa I need a review on this PR as well, it's similar to the other one except that it's forked out of 1.20 branch so that we can release this with Appium Desktop v1.20

app/main/appium.js
```diff
@@ -330,9 +330,15 @@ function connectClientMethodListener () {
 }
 
 const getCurrentSessions = _.debounce(async (evt, data) => {
-  const {host, port, path: appiumPath = '/wd/hub', ssl} = data;
+  const {host, port, path: appiumPath = '/wd/hub', ssl, username, accessKey} = data;
+  let res;
   try {
-    const res = await request(`http${ssl ? 's' : ''}://${host}:${port}${appiumPath}/sessions`);
+    if (username && accessKey) {
```
KazuCocoa, line 336: as same as https://github.com/appium/appium-desktop/pull/1691/files#r575626928

huzaifaiftikhar: Made these changes

KazuCocoa (approved):

KazuCocoa:
thanks!

PR #1694, opened 2021-02-12
feat: add sessionName in the attach to session dropdown

huzaifaiftikhar:
Once #1693 is merged users will be able to view the list of running sessions in the attach to session window. It is difficult to filter the sessions via the session-id, therefore, adding the `sessionName` (if present) to the list of `importantCaps`.

**Screenshot from the dev build**
<kbd>sessionName in attach to session list</kbd>

### DIFF
--- a/app/renderer/components/Session/AttachToSession.js
+++ b/app/renderer/components/Session/AttachToSession.js
@@ -4,6 +4,7 @@ import SessionCSS from './Session.css';
 import {
   ReloadOutlined
 } from '@ant-design/icons';
+import {ServerTypes} from '../../actions/Session';
 
 const FormItem = Form.Item;
 
@@ -12,6 +13,11 @@ function formatCaps (caps) {
   if (caps.automationName) {
     importantCaps.push(caps.automationName);
   }
+  return importantCaps.join(', ').trim();
+}
+
+function formatCapsBrowserstack (caps) {
+  let importantCaps = formatCaps(caps).split(', ');
   if (caps.sessionName) {
     importantCaps.push(caps.sessionName);
   }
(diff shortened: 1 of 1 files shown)


KazuCocoa:
What's the `sessionName`?
It seems the value is not a regular Appium stuff.

KazuCocoa:
btw, please sign the CLA

huzaifaiftikhar:
> What's the `sessionName`?
> It seems the value is not a regular Appium stuff.

Yes, it's not an appium capability, this capability is specific to BrowserStack by which we allow the user's to give a meaningful name to their sessions.
Would it possible to accept this PR as a feature request or do you think otherwise?

KazuCocoa:
Could it be possible to build your own custom AD for your platform with this change?

It would be better to define a new module or method to allow a vendor to customize the naming for them.
I don't check around the code well, but something like:

```js
function formatCaps (vendor, caps) {
    if (vendor === 'browserstack') {
        return formatCapsBrowserStack(caps)
    }
    // ...
}

function formatCapsBrowserStack (caps) {
    // do your own logic
}
```

Or extract the place as a different module.

huzaifaiftikhar:
@KazuCocoa can you re-review? I've created the method `formatCapsBrowserstack` and other cloud providers can change the formatting as per their choice using a similar approach.

app/renderer/components/Session/AttachToSession.js
```diff
@@ -15,14 +16,26 @@ function formatCaps (caps) {
...
+  }
+  return importantCaps.join(', ').trim();
+}
+
 export default class AttachToSession extends Component {
 
-  getSessionInfo (session) {
-    return `${session.id} — ${formatCaps(session.capabilities)}`;
+  getSessionInfo (session, serverType) {
+    if (serverType === ServerTypes.browserstack) {
```
KazuCocoa, line 30: Could you change here to `switch`?

app/renderer/components/Session/AttachToSession.js
```diff
@@ -15,14 +16,26 @@ function formatCaps (caps) {
   return importantCaps.join(', ').trim();
 }
 
+function formatCapsBrowserstack (caps) {
```
KazuCocoa, line 19: 👍

app/renderer/components/Session/AttachToSession.js
huzaifaiftikhar: Done

KazuCocoa (approved):

PR #1716, opened 2021-03-03
feat: override NodeJS' request method of the http(s) modules for adding a User-Agent header

### LINKED ISSUE #1708: Add a user-agent to all the requests being sent via Appium Desktop
## Appium or Appium Desktop?

You are reporting an issue at the Appium Desktop repository. Appium Desktop is a wrapper around Appium. If you are having trouble running tests, it is much more likely that the problem you are encountering is not a problem with Appium Desktop but with Appium. For that reason we require you to have tried your testcase with Appium CLI before reporting issues. Check the checkbox below to confirm that you have proven your issue does _not_ reproduce on Appium itself:

- [ ] I have verified that my issue does not occur with Appium and should be investigated as an Appium Desktop issue

## The problem
1. Currently, whenever a command/request is hit via Appium Desktop the user-agent is logged as `admc/wd/x.y.z`, this happens because we use the [admc](https://github.com/admc/wd) library for sending commands (excluding the master branch where we have started using [web2driver](https://github.com/projectxyzio/web2driver) as well)
2. Since admc can be used by other apps as well and is not just specific to Appium Desktop, therefore, the cloud providers are unable to identify by looking at the user-agent that whether the requests are coming from Appium Desktop or not.
3. As good practice and as per [RFC 7231](http://svn.tools.ietf.org/svn/wg/httpbis/specs/rfc7231.html#header.user-agent), an app should send a User-Agent, therefore, we should include a user-agent in all the request that we send via Appium Desktop. Ideally the user-agent should contain the name of the App and the version as well that makes it look something like this, `Appium Desktop - v1.20.2`
4. Found a similar issue - https://github.com/appium/appium/issues/4065

**Proposed Solution**
Since we use various libraries like admc/wd, request, etc. for sending HTTP requests, therefore, we can override the NodeJS' base modules such as [NET](https://nodejs.org/api/net.html) & [HTTP](https://nodejs.org/api/http.html) modules and extend these modules to add the extra header i.e. a user-agent for all the requests that we make via Appium Desktop.

## Environment

- I am running Appium Desktop version _<insert version here>_.
- I am on (pick one):
    - [ ] Mac
    - [ ] Windows
    - [ ] Linux

huzaifaiftikhar:
Closes #1708

### DIFF
--- a/app/override/override_http.js
+++ b/app/override/override_http.js
@@ -5,10 +5,9 @@ export const overrideHttpUserAgent = () => {
         originalHttpRequest = http.request;
 
   http.request = function (options, callback) {
-    let _options = options;
-    _options.headers['User-Agent'] = `${wdPackageJson.name}/${wdPackageJson.version} ${packageJson.name}/${packageJson.version}`;
+    options.headers['User-Agent'] = `${wdPackageJson.name}/${wdPackageJson.version} ${packageJson.name}/${packageJson.version}`;
 
-    return originalHttpRequest(_options, function (res) {
+    return originalHttpRequest(options, function (res) {
       if (callback) {
         callback(res);
       }
--- a/app/override/override_http.js
+++ b/app/override/override_http.js
@@ -1,0 +1,16 @@
+export function overrideHttpUserAgent () {
+  let packageJson = require('../../package');
+  let http = require('http');
+  let originalHttpRequest = http.request;
+
+  http.request = function (options, callback) {
+    let _options = options;
+    _options.headers['User-Agent'] = `${packageJson.name}/${packageJson.version}`;
+
+    return originalHttpRequest(_options, function (res) {
+      if (callback) {
+        callback(res);
+      }
+    });
+  };
+}
--- a/app/override/override_http.js
+++ b/app/override/override_http.js
@@ -1,11 +1,12 @@
-export function overrideHttpUserAgent () {
-  let packageJson = require('../../package');
-  let http = require('http');
-  let originalHttpRequest = http.request;
+export const overrideHttpUserAgent = () => {
+  const packageJson = require('../../package'),
+        wdPackageJson = require('../../node_modules/wd/package'),
+        http = require('http'),
+        originalHttpRequest = http.request;
 
   http.request = function (options, callback) {
     let _options = options;
-    _options.headers['User-Agent'] = `${packageJson.name}/${packageJson.version}`;
+    _options.headers['User-Agent'] = `${wdPackageJson.name}/${wdPackageJson.version} ${packageJson.name}/${packageJson.version}`;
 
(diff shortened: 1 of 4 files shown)


huzaifaiftikhar:
@KazuCocoa Can you please review?

app/override/override_http.js
```diff
@@ -0,0 +1,16 @@
+export function overrideHttpUserAgent () {
+  let packageJson = require('../../package');
+  let http = require('http');
+  let originalHttpRequest = http.request;
+
+  http.request = function (options, callback) {
+    let _options = options;
+    _options.headers['User-Agent'] = `${packageJson.name}/${packageJson.version}`;
```
KazuCocoa, line 8: Could you add `wd` and the version as well?
```
a = require('wd/package')
a.name # 'wd'
a.version # '1.13.0'
```

So, it will be like `wd/1.3.0 appium-desktop/1.20.2`

app/override/override_http.js
```diff
@@ -0,0 +1,16 @@
+export function overrideHttpUserAgent () {
+  let packageJson = require('../../package');
+  let http = require('http');
+  let originalHttpRequest = http.request;
```
KazuCocoa, line 4: The above can be const?

app/override/override_https.js
```diff
@@ -0,0 +1,16 @@
+export function overrideHttpsUserAgent () {
```
KazuCocoa, line 1: Same as http.

app/override/override_http.js
huzaifaiftikhar: Added the `wd` version to the User-Agent as well, on the 1.20 branch it looks like _"wd/1.13.0 appium-desktop/1.20.2-1"_

app/override/override_http.js
huzaifaiftikhar: Done

app/override/override_https.js
huzaifaiftikhar: Done

app/override/override_http.js
```diff
@@ -0,0 +1,17 @@
+export const overrideHttpUserAgent = () => {
+  const packageJson = require('../../package'),
+        wdPackageJson = require('../../node_modules/wd/package'),
+        http = require('http'),
+        originalHttpRequest = http.request;
+
+  http.request = function (options, callback) {
+    let _options = options;
```
KazuCocoa, line 8: Do we need this line?

huzaifaiftikhar: Yes, it was redundant, I've removed it.

KazuCocoa (approved):

KazuCocoa:
thanks :)

huzaifaiftikhar:
@KazuCocoa do you think we should create a similar PR against the master branch as well?

PR #1732, opened 2021-03-10
fix: request to remote server not sent in case of fresh installation

huzaifaiftikhar:
1. If Appium desktop is installed and we add a cloud provider then the "Attach to session" tab doesn't list any running sessions. This happens because the request to the remote server is never sent as the host, port & the path for the selected server type is not yet defined in the `session` variable. These values are populated only once we trigger a new session via AD for the newly added cloud provider, as it was only being set in the `NEW_SESSION_REQUESTED` event.
2. To fix the above issue I have created a method `getServerInfo` that returns the server information for the currently selected server type. This method either returns the values from the session variable or returns the default value.
3. The method `getServerInfo` is also used for both the `NEW_SESSION_REQUESTED` and `GET_SESSIONS_REQUESTED` events.
4. For reproducing on an already installed AD on macOS, run the following command `rm ~/Library/Application\ Support/appium-desktop/Settings` (⚠️ This will remove the saved settings so it might be better to take a backup of this file). After deleting restart AD -> add a cloud provider -> go to "Attach to Session" tab -> click on refresh button. Even if there are running sessions for your account on the remote server AD fails to list them.

### DIFF
--- a/app/renderer/actions/Session.js
+++ b/app/renderer/actions/Session.js
@@ -164,6 +164,142 @@ export function removeCapability (index) {
 }
 
 /**
+ * Get the server info for the current serverType
+ */
+export function getServerInfo (session) {
+  let host, port, username, accessKey, https, path, token;
+  switch (session.serverType) {
+    case ServerTypes.local:
+      host = session.server.local.hostname;
+      if (host === '0.0.0.0') {
+        // if we're on windows, we won't be able to connect directly to '0.0.0.0'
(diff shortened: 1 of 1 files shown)


huzaifaiftikhar:
@KazuCocoa Can you review this as well?

KazuCocoa:
Yes, I would do but haven't done yet.

KazuCocoa (approved):
Seems good, overall. Will merge this soon

PR #1753, opened 2021-03-30
ci: publish both apps with their own version scheme

jlipps:


jlipps:
closing, we're going to split apps into separate repos

PR #1906, opened 2021-09-16
fix hybrid element position determination

wswebcreation:
When switched to the webview of a hybrid app the elements on the screenshot were not aligned with the expected position.  See screenshot

 **Android:**
image

**iOS:**
image

This PR will fix that by determining the position of the first webview on in the hybrid app, see below

 **Android:**
image

**iOS:**
image

This will also fix the issue that Android hybrid apps got stuck when switching to the webview. The `nativeWebScreenshot` was needed to get a full screen screenshot instead of the ChromeDriver Webview screenshot

### DIFF
--- a/app/main/appium-method-handler.js
+++ b/app/main/appium-method-handler.js
@@ -191,7 +191,7 @@ export default class AppiumMethodHandler {
 
   async _getContextsSourceAndScreenshot () {
     let contexts, contextsError, currentContext, currentContextError, platformName,
-        source, sourceError, screenshot, screenshotError, statBarHeight, windowSize, windowSizeError;
+        source, sourceError, screenshot, screenshotError, statBarHeight, windowSize, windowSizeError, webViewPosition;
 
     if (!await this._hasContextsCommand()) {
       currentContext = null;
@@ -232,6 +232,15 @@ export default class AppiumMethodHandler {
       }
 
       if (currentContext !== NATIVE_APP) {
+        const isAndroid = _.toLower(platformName) === 'android';
+        try {
+          // Get the webview offset
+          const el = await this.driver.elementOrNull(
+            isAndroid ? 'xpath' : '-ios class chain',
+            isAndroid ? '//android.webkit.WebView' : '**/XCUIElementTypeWebView'
+          );
+          webViewPosition = await this.driver.getLocation(el.value);
+        } catch (ign) {}
         await this.driver.context(currentContext);
       }
       // End of note
@@ -242,9 +251,15 @@ export default class AppiumMethodHandler {
        */
       try {
         if (currentContext !== NATIVE_APP) {
-          const webviewStatusAddressBarHeight = await this.driver.execute(getWebviewStatusAddressBarHeight, [{platformName, statBarHeight}]);
-
-          await this.driver.execute(setHtmlElementAttributes, [{platformName, webviewStatusAddressBarHeight}]);
+          // Fallback if the webview position can't be determined,
+          // then do it based on the web context
+          if (!webViewPosition) {
+            webViewPosition = {
+              x: 0,
+              y: await this.driver.execute(getWebviewStatusAddressBarHeight, [{platformName, statBarHeight}]),
+            };
+          }
+          await this.driver.execute(setHtmlElementAttributes, [{platformName, webviewStatusAddressBarHeight: webViewPosition.y}]);
         }
       } catch (ign) {}
     }
--- a/app/main/appium-method-handler.js
+++ b/app/main/appium-method-handler.js
@@ -239,7 +239,7 @@ export default class AppiumMethodHandler {
             isAndroid ? 'xpath' : '-ios class chain',
             isAndroid ? '//android.webkit.WebView' : '**/XCUIElementTypeWebView'
           );
-          if (!webViewPosition) {
+          if (el) {
             webViewPosition = await this.driver.getLocation(el.value);
           }
         } catch (ign) {}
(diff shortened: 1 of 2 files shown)


app/main/appium-method-handler.js
```diff
@@ -242,9 +251,15 @@ export default class AppiumMethodHandler {
...
-          await this.driver.execute(setHtmlElementAttributes, [{platformName, webviewStatusAddressBarHeight}]);
+          // Fallback if the webview position can't be determined,
+          // then do it based on the web context
+          if (!webViewPosition) {
+            webViewPosition = {
+              x: 0,
+              y: await this.driver.execute(getWebviewStatusAddressBarHeight, [{platformName, statBarHeight}]),
+            };
+          }
+          await this.driver.execute(setHtmlElementAttributes, [{platformName, webviewStatusAddressBarHeight: webViewPosition.y}]);
```
mykola-mokhnach, line 262: previously we were passing `statBarHeight`. Is it expected that we change its name to `webviewStatusAddressBarHeight`?

wswebcreation: Hi @mykola-mokhnach 

Not sure what you mean, the arguments are exactly the same, only an extra if else check

mykola-mokhnach: sorry, got it now. Please ignore

app/main/appium-method-handler.js
```diff
@@ -232,6 +232,15 @@ export default class AppiumMethodHandler {
...
 
       if (currentContext !== NATIVE_APP) {
+        const isAndroid = _.toLower(platformName) === 'android';
+        try {
+          // Get the webview offset
+          const el = await this.driver.elementOrNull(
+            isAndroid ? 'xpath' : '-ios class chain',
+            isAndroid ? '//android.webkit.WebView' : '**/XCUIElementTypeWebView'
+          );
+          webViewPosition = await this.driver.getLocation(el.value);
```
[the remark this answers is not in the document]
wswebcreation: Fixed it

app/main/appium-method-handler.js
```diff
@@ -232,6 +232,17 @@ export default class AppiumMethodHandler {
...
 
       if (currentContext !== NATIVE_APP) {
+        const isAndroid = _.toLower(platformName) === 'android';
+        try {
+          // Get the webview offset
+          const el = await this.driver.elementOrNull(
+            isAndroid ? 'xpath' : '-ios class chain',
+            isAndroid ? '//android.webkit.WebView' : '**/XCUIElementTypeWebView'
+          );
+          if (!webViewPosition) {
```
mykola-mokhnach, line 242: sorry, I meant we need to check if `el` is null

wswebcreation: 🤦 , fixed it

app/renderer/actions/Session.js
```diff
@@ -802,29 +802,27 @@ export function setVisibleProviders () {
...
 function addCustomCaps (caps) {
-  const {browserName = '', platformName = ''} = caps;
-  const safariCustomCaps = {
-    // Add the includeSafariInWebviews for future HTML detection
-    includeSafariInWebviews: true,
-  };
-  const chromeCustomCaps = {
+  const {platformName = ''} = caps;
+  const androidCustomCaps = {
+    // @TODO: remove when this is defaulted in the newest Appium 1.8.x release
```
mykola-mokhnach, line 807: 1.8.x? must be a quite old one

mykola-mokhnach (approved):

app/renderer/actions/Session.js
wswebcreation: That was code I copied over, not sure if we still need it but didn't want to change that in this PR

jlipps:
Given that this is being targeted to the old AD repo, I don't see that it will ever be published. This PR should be opened against the inspector repo instead. (Unless @KazuCocoa is specifically publishing old versions of AD, which I think would be confusing now that we have fully made the switch).

KazuCocoa:
yea, the reason why I've maintained this branch was the inspector blocked https connection. That was already resolved in a recent update (HTTP libraries and webdriverio update). So it would be great to create the same change to the inspector repo.
And release the inspector first. (This PR itself can keep here)
if we still have some blocking issue, I'm okay to update this branch

afaik, after https://github.com/appium/appium-inspector/pull/118 , cloud vendors also can use appium-inspect on their env.
(some of them probably still need to update their capabilities. I already handle it in https://github.com/appium/appium-inspector/issues/128)

wswebcreation:
I totally understand @jlipps and @KazuCocoa 

But this is a fix for a customer who is using Sauce Labs which is currently not yet supported with the new one (or is it now). 
It would be great if we can make a patch release. I'll add this in the new repo next week or something, when I have some more time

jlipps:
@wswebcreation the new one will support sauce labs as soon as I merge the other PR. I think just need someone to sign the CLA for me to do that. so I would be able to publish that other one sooner than this one probably.

wswebcreation:
Which CLA do you mean @jlipps ?

I’ve build the app now on my local machine and send it to the customer so we’re good now

KazuCocoa:
I see.
I can create a patch version as 1.21.0-1.

1.22.0 version will be the separate inspector version, so the patch will reduce confusion, I think.

wswebcreation:
A patch would be great if that is possible

KazuCocoa:
https://github.com/appium/appium-desktop/releases/tag/v1.21.0-1 as a pre-release.

wswebcreation:
Thanks @KazuCocoa !!!

Issue #1995, opened 2022-03-28
Codesign has been expired for Win


KazuCocoa:
## Appium or Appium Desktop?

Codesign for Win failed as https://dev.azure.com/AppiumCI/Appium%20CI/_build/results?buildId=20481&view=logs&j=f87c6499-3401-572f-ac24-cec18d4bfbce&t=4032fcf8-7e70-54e0-2d0a-53cd538a71f4

I guess the motivation about it is the same as macOS. So, on Win 10 in my local, the below security permission appeared.
But once I allowed it, it had no issue.

<img width="488" alt="Screen Shot 2022-03-27 at 23 48 19" src="https://user-images.githubusercontent.com/5511591/160341761-81486004-0619-4362-acd5-526664572ea6.png">

If 1.22.3 with no code-signature (by JSFoundation?) works, probably we completely can skip the sign since we will not publish it via their stores.

## The problem

_Replace this section with a description of your issue_

## Environment

- I am running Appium Desktop version _<insert version here>_.
- I am on (pick one):
    - [ ] Mac
    - [ ] Windows
    - [ ] Linux

suleytron:
The same issue seems to be present on Mac as well.

eoghanp95:
Also running into this issue on Mac

PR #1996, opened 2022-03-28
chore: comment out CSC

KazuCocoa:
https://dev.azure.com/AppiumCI/Appium%20CI/_build/results?buildId=20481&view=logs&j=f87c6499-3401-572f-ac24-cec18d4bfbce&t=4032fcf8-7e70-54e0-2d0a-53cd538a71f4 failed in Win build as code signature has been expired.

I don't know where sets the signature. Probably the below section in the Azure??? I don't know the exact value neighter.
Screen Shot 2022-03-28 at 0 05 20

In my local Win app (Win 10 env though) worked without the sign. So, a module by `npx electron-builder build -w --ia32 --x64` worked.

So, I wonder we can remove `CSC_LINK` for this purpose since we won't publish these apps to their online store. Or we probably can ask JSFoundation...? @jlipps 

https://github.com/appium/appium-desktop/releases/tag/v1.22.3 's Win modules are built by `npx electron-builder build -w --ia32 --x64 --publish never` on my local without such `CSC_*`. We can merge this PR if we do not get any errors by such signatures in the version.

----

https://github.com/appium/appium-desktop/issues/1995

### DIFF
--- a/ci-jobs/templates/package.yml
+++ b/ci-jobs/templates/package.yml
@@ -32,5 +32,5 @@ jobs:
       displayName: Bundle
       env:
         GH_TOKEN: $(GH_TOKEN)
-        CSC_LINK: $(CSC_LINK)
-        CSC_KEY_PASSWORD: $(CSC_KEY_PASSWORD)
+        # CSC_LINK: $(CSC_LINK)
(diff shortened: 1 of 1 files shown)


jlipps:
👍  we can also ask the OpenJSF for a new signature.

PR #2005, opened 2022-04-02
chore: update electron 9

KazuCocoa:
https://www.electronjs.org/blog/electron-10-0

This change worked.


KazuCocoa:
v8 and v9 seem good.
I'll merge this if CI passes. Then, I'll update package.json with nodejs 14 base.
Then, will try to update the electron to 11, that is NodeJS 14 based one.

PR #2009, opened 2022-04-04
chore: update electron to v8

KazuCocoa:
~~Maybe this will pass since a couple of days worked...~~
=> ok, at least electron v8 passed the ci


jlipps (approved):
looks good to me if you can build it locally and it works! i think i can't build this locally right now for M1 reasons.

PR #2106, opened 2022-09-15
fix: change stop-server icon

depapp:
hi all,

since the icon for `Stop Server` is using `PauseOutlined`, I think it's more make sense to use `StopOutlined` instead.

- after this PR
Screen Shot 2022-09-15 at 11 26 31

- before this PR
Screen Shot 2022-09-15 at 13 29 41


jlipps:
hmm, not really a fan of that icon either, what  about finding a square stop button?

depapp:
> hmm, not really a fan of that icon either, what about finding a square stop button?

what do you think about this square icon @jlipps ?
image

jlipps:
I think it looks good but it should be filled in so it looks like a stop button and not an empty square/icon/box.

